Issue #40, opened 2015-11-04
Improve duration accuracy


seabaylea:
The duration value in events is currently millisecond based, generated using Date.now(). This isn't particularly useful for short (sub-millisecond) events, so we should move to using a more accurate timer, eg. process.hrtime();

tunniclm:
Is this issue complete with the merge of #42?

seabaylea:
Yup. Closing.

PR #42, opened 2015-11-06
Convert durations to using process.hrtime() for improved accuracy

seabaylea:
Issue #40 covers improving duration accuracy beyond millisecond granularity. This is done by moving to the use of process.hrtime() for the measurements, and converting to results to still use milliseconds as the result.


seabaylea:
@tunniclm Can you review, provide feedback and merge as necessary?

tunniclm:
LGTM

Issue #43, opened 2015-11-06
Set sensible default for minClockStack


seabaylea:
The 'request' capability can optionally create stack traces for any monitoring event in the request that exceeds a given duration. The default is currently set to 0ms, so stack traces are created for every monitoring event, which is very expensive.

This either needs to be disabled by default, or set to a reasonable default threshold value.

seabaylea:
The proposal would be to set a default of '-1' for disabled, and allow users to configure if they want stack generation.

seabaylea:
Delivered by PR #47

PR #47, opened 2015-11-09
Disable stack traces on request events by default

seabaylea:
Issue #43 cover the need to set a sensible default for stack trace generation in request tracking. This should do that by allowing a value of '-1' to be used to disable stack trace generation, and to make that the default.

Additionally I've removed the minCpuStack check - until we implement CPU usage tracking per-request, the check is superfluous.


seabaylea:
@tunniclm Can you review, provide feedback and merge as necessary?

tunniclm:
LGTM. When these configuration options are ready to be exposed generally, they will need to be documented.

Issue #46, opened 2015-11-09
Expose patched function name in aspects framework


seabaylea:
The aspects patching framework allows you to specify an array of function names to patch. eg. for aspect.around:

    var functionNames = [];
    aspect.around(target.prototype, functionNames,
        function(target, args) {},
        function(target. args, rc) {}
    );

This means you can patch multiple functions in the same way with the same code, but there's no way of knowing the name of the actual function that is running. This is useful if you want to add it to the raised monitoring event.

Passing the function name through to the function callbacks as an additional (last) parameter provides access to the name of the function without breaking existing code, eg:

    var functionNames = [];
    aspect.around(target.prototype, functionNames,
        function(target, args, funcName) {},
        function(target. args, rc, funcName) {}
    );

seabaylea:
@tunniclm Any view on the addition of the patched function name as a new last parameter passed to the supplied before and after callbacks?

Note that this only applies to aspect.before(), aspect.after() and aspect.around(). The aspect.aroundCallback() call doesn't take a function name - you should already have that from a previous call to aspect.before() or aspect.around().

tunniclm:
@seabaylea This approach seems reasonable to me, adding as the last argument should prevent breakage of existing usage code.

seabaylea:
Closing with PR #49

PR #49, opened 2015-11-10
Expose patched method/function name in aspects API

seabaylea:
This provides the "methodName" that's been patched to the function callbacks to the aspects.before(), aspects.after() and aspects.around() calls so that its possible to know which function is being patched if an array of function names was provided.

Its passed as the new last parameter to avoid affecting existing code that uses the aspects calls.

### DIFF
--- a/lib/aspect.js
+++ b/lib/aspect.js
@@ -37,19 +37,19 @@ function aspect(type) {
 			var newFunc;
 			if(type == 'before') {
 				newFunc = function() {
-					hookBefore(this, arguments);
+					hookBefore(this, arguments, methodName);
 					return existing.apply(this, arguments);
 				};
 			}
 			else if(type == "around") {
 				newFunc = function() {
-					hookBefore(this, arguments);
+					hookBefore(this, arguments, methodName);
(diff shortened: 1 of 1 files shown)


seabaylea:
@tunniclm Can you review this as the fix for issue #46

lib/aspect.js
```diff
@@ -85,7 +85,7 @@ exports.aroundCallback = function(args, hookBefore, hookAfter) {
 		var ret = orig.apply(this, arguments);
 
 		if(hookAfter) {
-			hookAfter(this, arguments);
+			hookAfter(this, arguments, ret);
```
tunniclm: @seabaylea Is this change another fix? It doesn't look related to the method name change, unless I'm missing something?

seabaylea:
Yeah, its a "small fix" which didn't look like it really deserved its own issue... hookAfter should have three parameters: this, arguments and return value. For some reason that call was missing the return value.

tunniclm:
OK, LGTM then. Are you ready for it to be merged?

seabaylea:
Yup - it should be good to go.

Issue #48, opened 2015-11-10
HTTP event URLs can be incorrect when using Express routing


seabaylea:
When using Express with router level middleware, the req.url value at the start of the request (which the user makes requests of) is modified to remove the context root / mount URL before the request completes.

As the HTTP event currently uses the req.url value that's present at the end of the request, it doesn't match the external URL the request is made against.

Using the actual hosted URL makes more sense, especially as the filters configuration can be used to modify the URL to something else if required.

seabaylea:
Resolved by PR #50

PR #50, opened 2015-11-10
Ensure the external URL is used in HTTP events

seabaylea:
This PR covers issue #48 by taking the value for traceUrl, which is based off the incoming URL, and reusing that as the URL for the raised event. This is the correct behaviour as the traceUrl is result of url.parse().url after any transform/grouping has occurred - which is what we want to raise the event against.

As a happy by-product, this change has removed an already redundant call to if (traceUrl !== ''), and removed a call to url.parse which is no longer required.


tunniclm:
LGTM, ready for merge?

seabaylea:
Yup - should be good to go.

Issue #51, opened 2015-11-10
Remove unnecessary calls to check if process.version > 0.8


seabaylea:
In `lib/aspects.js` the following check is done in two places:
```js
if(process.version.split('.')[0] == 'v0' && process.version.split('.')[1] < 8){}
```

This isn't needed because Node Application Metrics doesn't versions before 0.10 (there is an install time check for this), so the check is redundant.

seabaylea:
Resolved by PR #52

PR #52, opened 2015-11-10
Remove checks for Node.js version < 0.8

seabaylea:
Issue #51 was raised to remove unnecessary calls to check if the Node.js version was less that 0.8. This does just that.

Providing the JIT doesn't optimize out the checks already, this should have a decent effect on performance as the check are done for every end probe.

@tunniclm can you provide a quick review?


tunniclm:
LGTM, ready to merge?

seabaylea:
Yup, should be good to go.

PR #56, opened 2015-11-12
Remove millisecond duration value

seabaylea:
This cleans up timer.js.

The `duration` value was new in issue #40 and added in case changing `timeDelta` to a higher precision caused an issue. It looks like that's not the case, so we can remove `duration` and assign the value to the original `timeDelta` field.

I've additionally removed `clockTime()`, `clockCpuTime()` and `cpuTimeDeltaInMs()` as they're not used for anything.


seabaylea:
@tunniclm could you review?

tunniclm:
LGTM, ready to merge?

seabaylea:
Yup, should be good to go.

Issue #57, opened 2015-11-13
Improve performance of HttpProbe.filterUrl()


seabaylea:
Profiling the overhead of appmetrics with the AcmeAir benchmark, and visualizing the data in the Health Center client, shows that 2.47% of CPU is being spent executing `HttpProbe.filterUrl()` and the functions it calls:
![image](https://cloud.githubusercontent.com/assets/11974522/11142893/8eea3a8e-89e8-11e5-9f48-7f2c0b596c2e.png)

`HttpProbe.filterUrl()` itself is only using 0.021% (Self %) and the functions it calls account for the rest. The called methods shows that almost all of this time is spent executing `Url.parse()`.

Whilst `Url.parse()` can parse the full range of URL types and break it down into constituent parts, the use case here is much more limited - we're just taking the `req.url` value from the HTTP request (which don't contain hostname/port etc), and stripping off any query on the end.

This can be done much more cheaply using a custom implementation rather than using `Url.parse()`, ie. stripping the URL at the first instance of '#' or '?'.

seabaylea:
Using a simple `indexOf()` / `substring()` implementation like the following:

```js
var parse = function (url) {
	['?','#'].forEach(function (separator)  {
		var index = url.indexOf(separator);
		if (index !== -1) url = url.substring(0, index);
	});
	return url;
};
```
Shows a 3x improvement over `Url.parse()` for a URL with containing a '?' query. This means that the change should result in a ~1.6% reduction in the performance overhead of appmetrics.

seabaylea:
Resolved by PR #59

seabaylea:
Latest data has the cost of `HttpProbe.filterUrl()` and the functions it calls down to 0.24%

PR #59, opened 2015-11-13
Improve performance of HttpProbe.filterUrl()

seabaylea:
Issue #57 covers the improvement of HttpProbe.filterUrl() performance. This achieves that by replacing the use of Url.parse() with a custom implementation that relies on the fact that we're not passing in a full URL, but only needs to have any query stripped of the end of the URL.

@tunniclm could you review?

### DIFF
--- a/probes/http-probe.js
+++ b/probes/http-probe.js
@@ -58,10 +58,21 @@ HttpProbe.prototype.attach = function(name, target, am) {
 };
 
 /*
+ * Custom req.url parser that strips out any trailing query
+ */
+var parse = function (url) {
+	['?','#'].forEach(function (separator)  {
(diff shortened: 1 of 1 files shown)


tunniclm:
I quickly checked https://nodejs.org/api/http.html#http_http_incomingmessage and the URI spec and it looks like the `request.url` would always be relative to the server root, which makes the simplification seem valid. Also, from my reading of the URI spec the first plain occurrence of `?` and `#` should signify the start of the query/fragment parts of the URI (any earlier occurrence for a different meaning would have to be % encoded).
So...LGTM, ready to merge?

seabaylea:
Yup, should be good to go.

Issue #61, opened 2015-11-23
Global variables in probes can cause requests to be overwritten.


hhellyeribm:
Several probes are storing request data and timer data in global variables which means that multiple events can end up overwriting each others data.

seabaylea:
It looks like there's some mistaken use of global rather than local variables which means that if you have several events fired (particularly of the same type) in very quick succession than the data can be "merged".

This doesn't appear to affect HTTP too badly, but if there's multiple MySQL or MongoDB calls in the same HTTP request then some merging happens.

This is critical to fix as it relates to the correctness of the data.

seabaylea:
I haven't found any issues with this in integration testing, so closing.

PR #62, opened 2015-11-23
Global variables in probes can cause requests to be overwritten.

hhellyeribm:
Fix for issue #61 

Correct the use of globals inside the probes and change parameter passing. Update existing probes to and aspects to pass context so that they can maintain state from the initial function call until callbacks are executed without the use of global variables.
Refactor and simplify aspect.js to be clearer.
Allow shared timers between metrics and request probes to prevent duplicate timer objects being created.

### DIFF
--- a/index.js
+++ b/index.js
@@ -53,7 +53,9 @@ files.forEach(function (fileName) {
  * Patch the module require function to run the probe attach function
  * for any matching module. This loads the monitoring probes into the modules
  */
-aspect.after(module.__proto__, 'require', function(obj, args, ret) {
+var data = {};
+
+aspect.after(module.__proto__, 'require', data, function(obj, methodName, args, context, ret) {
 	for (var i = 0; i < probes.length; i++) {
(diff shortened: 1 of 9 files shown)


seabaylea:
Ok, so it looks like we now have a `context` object that's created in aspects.js and passed around as a parameter (removing the global variable issue). The probes can then use this to store a common `timer` object and any `request` object, to ensure the we stop the correct timer and request.

The use of a dummy/empty `data` object in index.js where there isn't a context object to pass is hacky - it should be removed and replaced with proper parameter parsing in aspect.js, but I don't think it should hold up delivery of the pull request.

seabaylea:
The rest looks good to me. Is this ready to be merged?

hhellyeribm:
Yes, should be ready to go.

seabaylea:
Ok - I'll merge it in. I'll keep the issue open until I've done some further integration testing.

Issue #69, opened 2015-11-26
Tracing sets incorrect context data


seabaylea:
When trace events are present in the request data, rather than having a context object being set for the event containing the arguments, eg:
```
{ arg0: 'customerSession',
  arg1: '03b05742-c43a-46c5-a921-d5ce6bc75157',
  arg2: 'function <anonymous>' }
```
what's actually present is a function, eg:
```
[Function]
```

It looks like the issue is that rather than calling the function to create the context data in `probes\trace-probe.js` and adding that to the request as the context, we're passing the function itself.

seabaylea:
Fixed by PR #70

PR #70, opened 2015-11-26
Pass trace context as an object

seabaylea:
This is the fix for issue #69 

Rather than passing the function that generates the context to `req.stop()` we should be passing the object returned from the object. This is a simple matter of just adding `()`!


seabaylea:
@tunniclm could you review?

tunniclm:
LGTM, merging.

Issue #71, opened 2015-11-26
Tracing of constructors can cause constructed object not to be returned


seabaylea:
When we run tracing of a constructor function, eg:

```js
myObj = new mod.Construct(data);
```

the instrumentation that's injected to generate the tracing fails to return the constructed object, so `myObj` is undefined.

seabaylea:
Fixed by PR #73

PR #73, opened 2015-11-26
Pass 'this'parameter to apply call

seabaylea:
This is the fix for issue #71 which causes constructed objects not to be returned if the constructor has been instrumented for trace.

The issue was that we were failing to pass `this` through to the `apply()` call for the instrumented method, which has now been resolved.


seabaylea:
@tunniclm could you review?

tunniclm:
It looks like this area of code could do with some attention to clearly define all the cases it needs to handle (with some nice testcases). I am suspicious that we aren't handling every cases correctly.

That said, this change is a clear improvement so I'm happy to merge.

PR #77, opened 2015-11-27
Agent version incorrect

tobespc:
I've modified the agent version to be appmetrics.version and also added agentcore.version to make displaying of the data in the health center client clearer when monitoring node. 

Fix for issues  #78


seabaylea:
LGTM so merging.

@tobespc its relatively minor, but in future would it be possible for you to submit the PR as a single commit (ie. without the merge commit)?

Issue #68, opened 2015-11-26
Add support for Redis monitoring


hhellyeribm:
Add appmetrics probe support for the redis npm module.

The data produced would be:

# Event: 'redis'

Emitted when a Redis client sends commands to a Redis server emit a 'redis' event and create a request.

## Event:
* __data__ (Object) the data from the redis event:
  * __time__ (Number) the time in milliseconds when the Redis event occurred. This can be converted to a Date using new Date(data.time)
  * __args__ (Array) A truncated version of the arguments to the Redis command. No more than 10 arguments are returned and string arguments are truncated. Callbacks are not included in the arguments.
  * __cmd__ (String) the Redis command sent to the server or 'batch.exec'/'multi.exec' for groups of command sent using batch/multi calls.
  * __duration__ (Number) the time taken in milliseconds.

## Request:
* __type: 'redis'__ - a request with type: 'redis' is started when a call is made to Redis.
* __root request: false__ - Redis requests are started by the running nodejs program not triggered by external events so they are children of a request triggered from outside.
* __context__ - the context object will contain:
  * __cmd (String)__ the Redis command sent to the server or 'batch.exec'/'multi.exec' for groups of command sent using batch/multi calls.
  * __args:  (Array)__ A truncated version of the arguments to the Redis command. No more than 10 arguments are returned and string arguments are truncated. Callbacks are not included in the arguments.

## Testcase:

* Connect to redis server and runs several basic commands.
* Create batch and multi objects and runs those.
* Create a publisher and subscriber test.
* Count the correct number of events have been received and that call backs passed to those calls were run correctly.

seabaylea:
Closding based on PR#79

PR #79, opened 2015-11-30
Implmenentation of appmetrics probe for the redis npm module.

hhellyeribm:
Add support for monitoring of redis ( for issue #68 )
Provides monitoring for redis commands issued via the client api or via batch/multi objects.
Test case verifies we get the expected number of events and that callbacks are still invoked.
Update README.md to include a description of the redis probe.


hhellyeribm:
Updated pull request. Removed arguments lists and just pass the command being run in. This simplifies the initial drop and data can be added back in if needed based on user feedback. (The batch and multi command arguments are complicated and feedback on what to provide for those would be useful to have first.)

seabaylea:
LGTM. Merging.

Issue #80, opened 2015-11-30
MySQL and MongoDB probes set request context twice.


hhellyeribm:
The mysql and mongo probes generate the context data for requests in requestStart() and requestEnd(). This only needs to be done once per request, the second context object will overwrite the first and both generate the same data in both cases.
As this involves making the relatively expensive call to JSON.stringify() we should remove the extra calls to improve performance.

PR #81, opened 2015-11-30
Remove the call to set the context in requestStart for the mysql and mongo probes.

hhellyeribm:
This patch should resolve issue #80 

The context is set in requestEnd and doesn't need to be set twice. This saves the cost of one of the JSON.stringify calls.
I removed the one in requestStart as that was being done while the timer was running. The call in requestEnd occurs once the timer has stopped so won't be counted towards the request time.


seabaylea:
Yeah, it looks like the `setContext()` calls are no longer required. I've done some integration testing and can't find a scenario where we aren't collecting the necessary context data for MongoDB or MySQL.

Merging.

Issue #74, opened 2015-11-26
Bad physical_used value from memory event on Mac OSX 32-bit


tunniclm:
Testcase:

```js
var assert = require('assert');
require('appmetrics').monitor().on('memory', function(data) {
  assert(data.physical_used === -1 || data.physical_used >= 0,
           'Memory data message does not contain a valid used physical memory ('
            + data.physical_used + ')');
});
setTimeout(function() {}, 3000);
```

Output:

```
assert.js:89
  throw new assert.AssertionError({
  ^
AssertionError: Memory data message does not contain a valid used physical memory (-4115607553)
    at API.<anonymous> ([eval]:1:80)
    at emitOne (events.js:77:13)
    at API.emit (events.js:169:7)
    at API.formatMemory (.../test/node_modules/appmetrics/appmetrics-api.js:125:14)
    at API.raiseEvent (.../test/node_modules/appmetrics/appmetrics-api.js:40:17)
    at events (.../test/node_modules/appmetrics/appmetrics-api.js:178:9)
```

PR #89, opened 2015-12-02
Fix physical used garbage value issue #74

tunniclm:
Looks like the bug is with the api only which generates an extra value (used physical memory) calculated by subtracting two others (total physical memory - free physical memory). These values are allowed to be `-1` to indicate the value is not available for the platform, however, the api code did not check for this and so would calculate the value incorrectly.

This pull request should fix issue #74.


tunniclm:
@seabaylea Can you take a quick look? If this looks good to you, then I will merge.

seabaylea:
Tthis looks good as it handles the case where physical_used is -1 (ie, not implemented). Merging.

Issue #53, opened 2015-11-10
Add max heap to node environment data source


tunniclm:
Provide the V8 max heap value in the node environment plugin data source. Since the max heap size will not change it should be fine to provide it once in the node environment data.

tunniclm:
It is simple enough to expose the heap_size_limit() value that can be obtained from the V8 HeapStatistics API. However, investigation shows that this value is (4*max_semi_space_size + max_old_generation_size) which doesn't actually represent the maximum size of the heaps.

There is one old generation space and the young generation is made up of 2 semi spaces. The heap_size_limit() gives the amount of address space **reserved** for these, which includes extra alignment address space for the semi spaces -- V8 reserves double the semi space size for this purpose.

I think what we actually need to properly represent the max heap size is (2*max_semi_space_size + max_old_generation_size). I cannot find a simple way to calculate this, since neither the max_semi_space_size or max_old_generation_size do not seem to be exposed by the V8 public API.

tunniclm:
Also, it would also be interesting to know the max_old_generation_size as a value independent of the calculation of the max heap size.

Unfortunately, the new HeapSpaceStatistics API does not provide maximum sizes for the spaces either.

tunniclm:
It is possible to calculate the values for max old space and max semi spaces sizes based on replicating the behaviour of V8. I have raised a pull request with an implementation that appears to work for the versions of V8 current for Node v0.10, v0.12, v4.

It adds the following keys to the environment object/message:
```
heap.size.limit=<bytes>
max.old.space.size=<bytes>
max.semi.space.size=<bytes>
```
`heap.size.limit` is the heap_size_limit value from the HeapStatistics API (in bytes)
`max.old.space.size` is the calculated value of the old generation max size (in bytes)
`max.semi.space.size` is the calculated value of **half** the young generation max size (in bytes)

seabaylea:
Resolved by PR #90

seabaylea:
Note that we also added the following during the review cycle:
`max.heap.size=<bytes>`
which represents the maximum heap size value that the GC used value can be compared to.

PR #90, opened 2015-12-02
Add max heap sizes to environment data for issue #53

tunniclm:
The max semi space and max old space sizes are not exposed
by the V8 api, so we calculate them from the provided command
line arguments, using default values if no argument is supplied.

Note: these changes are closely coupled with the internal
      implementation of V8 and are brittle to changes.
      Hopefully the required data will be exposed in a future
      V8 version and these values can then be taken directly.

### DIFF
--- a/src/plugins/node/env/nodeenvplugin.cpp
+++ b/src/plugins/node/env/nodeenvplugin.cpp
@@ -51,6 +51,7 @@ namespace plugin {
 	std::string commandLineArguments;
 	size_t maxOldSpaceSizeGuess;
 	size_t maxSemiSpaceSizeGuess;
+	size_t maxHeapSizeGuess;
 	size_t heapSizeLimit;
 }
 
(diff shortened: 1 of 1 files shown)


tunniclm:
@seabaylea Can you review please? If it looks good to you, then I'm happy to merge.

NOTE: No README update for this at present, since we don't document the content of the environment object yet.

seabaylea:
One of the key use cases for this data is to be able to calculate the JavaScript heap usage by live data as a percentage of its maximum possible size. This means being able to compare the `gc.used` value to the largest value that `gc.used` could potentially grow to (ignoring fragmentation etc).

The `heap.size.limit`, `max.semi.space.size` and `max.old.space.size` values are useful in themselves, but we should also add a "maximum possible `gc.used`" value.

tunniclm:
@seabaylea I've added `max.heap.size` which is `2*max.semi.space.size + max.old.space.size` and should fulfill the requirement. Feel free to merge if it look good.

seabaylea:
LGTM. Merging

Issue #38, opened 2015-10-25
Add support for memcached monitoring


seabaylea:
Memcached is one of the more popular datastores used by Node.js applications. The 'memcached' module currently has > 150K downloads a month, so it seems reasonable to provide support for monitoring the module.

seabaylea:
I've prototyped support in my personal fork (https://github.com/seabaylea/appmetrics). This produces the following data:

### Event: 'memcached'
Emitted when a data is stored to or retrieved from Memcached using the `memcached` module.
* `data` (Object) the data from the memcached event:
    * `time` (Number) the milliseconds when the memcached event occurred. This can be converted to a Date using `new Date(data.time)`
    * `method` (String) whether the event was for a data `set` or `get`.
    * `key` (String) the key associated with the data.
    * `duration` (Number) the time taken for the get or set of the memcached data to occur.

This is also included in request tracking, but neither a set or a get will start a 'request'.

Any feedback anyone has is appreciated.

seabaylea:
The prototype covers calls to the following functions:
    `get`, `gets`, `getMulti`, `set`, `replace`, `add`, `cas`, `append`, `prepend`, `increment`, `decrement`, `incr`, `decr`, `del`, `delete`

I've not added support for `version`, `flush`, `samples`, `slabs` and `items` as they seem to be more one-off/non-performance critical calls.

This can easily be added if there's a requirement though.

seabaylea:
Closing based on PR #94

PR #94, opened 2015-12-03
Add support for Memcached monitoring

seabaylea:
This covers issue #38.

The probe currently only collects basic function/method and key information. It can be extended with further context data based if we receive requests for useful data to be added.


tobespc:
there is a console.log(methodArgs[0]) in the file, should that be in there ?

seabaylea:
Thanks @tobespc ! I can't believe I left that in ';-) Now removed.

Issue #96, opened 2015-12-04
Centralise probeAttached check into probe handler


seabaylea:
Currently its expected that each of the probes will set and check whether the `__ddProbeAttached__` value is set on a module to denote that its already been probed.

This should be moved to the probe handler, both to remove complexity from the probes and reduce the risk that a probe implementor will fail to do the check properly.

Probes will still need to set the value themselves as probes that return constructors need to run every time and therefore should not set the value at all.

seabaylea:
Closing based on PR #97

PR #97, opened 2015-12-04
Carry out probe attached check centrally

seabaylea:
This is the PR for issue #96 

It moves the check for `__ddProbeAttached__` into index.js were we run the probes against the modules as their required.

Modules that return constructors are a special case. As the socket.io probe does that, there's been some fixup to that probe. The socket.io probe doesn't set `__ddProbeAttached__` to ensure that the constructor is probed each time, and the probes to the module prototypes are wrapped in a module specific `__prototypeProbeAttached__` check to make sure those are only probed once.

### DIFF
--- a/index.js
+++ b/index.js
@@ -56,15 +56,19 @@ files.forEach(function (fileName) {
 var data = {};
 
 aspect.after(module.__proto__, 'require', data, function(obj, methodName, args, context, ret) {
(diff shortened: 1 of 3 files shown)


hhellyeribm:
I've reviewed the patch in seabaylea:attached-check - it seems to resolve this problem correctly.

Issue #87, opened 2015-12-02
cannot enable('trace') for ghost v0.7.1 application


xiaoshimimi:
If we enable trace, with appmetrics, and run ghost v0.7.1 application, it through exception repeately: 
```
Unhandled rejection TypeError: Cannot read property 'clients' of undefined
    at renderIndex (/opt/node-v0.10.33/sample/ghost_v0.7.1/core/server/controllers/admin.js:22:72)
    at tryCatcher (/opt/node-v0.10.33/sample/ghost_v0.7.1/node_modules/bluebird/js/main/util.js:24:31)
    at Promise._settlePromiseFromHandler (/opt/node-v0.10.33/sample/ghost_v0.7.1/node_modules/bluebird/js/main/promise.js:454:31)
    at Promise._settlePromiseAt (/opt/node-v0.10.33/sample/ghost_v0.7.1/node_modules/bluebird/js/main/promise.js:530:18)
    at Promise.b (domain.js:183:18)
    at Async._drainQueue (/opt/node-v0.10.33/sample/ghost_v0.7.1/node_modules/bluebird/js/main/async.js:182:12)
    at Async._drainQueues (/opt/node-v0.10.33/sample/ghost_v0.7.1/node_modules/bluebird/js/main/async.js:187:10)
    at Async.drainQueues (/opt/node-v0.10.33/sample/ghost_v0.7.1/node_modules/bluebird/js/main/async.js:15:14)
    at process._tickDomainCallback (node.js:463:13)
```
Ghost v0.7.1: 
https://ghost.org/download/

To run ghost application, you must change "127.0.0.1" or "localhost" to your IP, otherwise it cannot be accessed outside this machine.

seabaylea:
I've reproduced this. We'll start working on a fix.

hhellyeribm:
@xiaoshimimi - I've pushed a fix that I believe resolves this issue. (It works for me!) Would you be able to verify it in your environment?

tunniclm:
FYI, #105 has been included in v1.0.4

xiaoshimimi:
The major problem is fixed.  But there is two small problem, which will not broke the application's function, but error messages throw out:

Like
```
ERROR: Object.keys called on non-object
 Checking for updates failed, your blog will continue to function.
 If you get this error repeatedly, please seek help from https://ghost.org/forum.
 TypeError: Object.keys called on non-object
    at Function.keys (native)
[...]

PR #105, opened 2015-12-08
Add missing return from callback.

hhellyeribm:
One of the cases where a call back function is wrapped by trace was missing a return statement.
This fixes issue #87 - cannot enable('trace') for ghost v0.7.1 application


seabaylea:
This is the PR for Issue #87

Issue #110, opened 2015-12-10
Add probe data is set made available via MQTT


tobespc:
when monitoring the appmetrics data with the Health Center client, none of the probe information is being transmitted.  Need to have the appmetrics agent enhanced to transmit probe data

hhellyeribm:
Changed the label - reading the above description this sounds like a bug.

seabaylea:
This isn't a bug - it's a deliberate design decision. The only current consumer of the MQTT connection - the Health Centre plug-in for the Eclipse IDE - does not utilise the data, therefore there was little point in paying the performance cost of transmitting the data.

Adding the JS level data into the set transmitted via MQTT makes sense, if/when a consumer (i.e., Health Centre) wants access to it. It sounds like this is a request for that to happen.

At the same time, I think we should be looking at removing the creation of the MQTT connection by default where 'appmetrics' is required into the application rather than launched using the `node-hc` CLI . That would happen under a different issue/enhancement.

PR #111, opened 2015-12-10
Add support for com.ibm.diagnostics.healthcenter.probes.transmit

hhellyeribm:
This patch adds a new property to the appmetrics.properties file com.ibm.diagnostics.healthcenter.probes.transmit which can be true or false.
If set to true probe data will be sent to the monitoring client.
If set to false the data will not be sent. false is the default.

I have added a new native method (nativeEmitCheck) that reports whether data should be transmitted. This property is checked at startup and it's value cached and the native method is only called once, at start up, with the true/false value then stored in index.js. This prevents repeated calls from Javascript to native code.

In the future the property could be extended to be a comma separated list of probes to send data for. (With true enabling all probes as now and false disabling all probes.) Changing this would not break users who wanted to continue setting true or false.
nativeEmitCheck() could be extended to take a probe name and return whether that probe was enabled. (If neccessary the no-args form could still report whether all probes were globally enabled or not.) nativeEmitCheck() is not part of the public API so could be safely changed or replaced if necessary.

This is for issue #110

### DIFF
--- a/appmetrics.properties
+++ b/appmetrics.properties
@@ -19,4 +19,4 @@ com.ibm.diagnostics.healthcenter.data.profiling=off
 
 # Allow probes to transmit data to the client: true | false
 # Defaults to false - probe data not sent.
-#com.ibm.diagnostics.healthcenter.probes.transmit=false
+#appmetrics.probes.transmit=false
--- a/appmetrics.properties
+++ b/appmetrics.properties
@@ -16,3 +16,7 @@ com.ibm.diagnostics.healthcenter.logging.plugins=info
 
 # Start with method profiling enabled/disabled: on | off
 com.ibm.diagnostics.healthcenter.data.profiling=off
+
+# Allow probes to transmit data to the client: true | false
+# Defaults to false - probe data not sent.
+#com.ibm.diagnostics.healthcenter.probes.transmit=false
--- a/index.js
+++ b/index.js
@@ -29,6 +29,10 @@ agent.start();
 
 var hcAPI = require("./appmetrics-api.js");
 
+// This is set by the property com.ibm.diagnostics.healthcenter.probes.transmit
+// at startup. Cache the value so we only make one native call.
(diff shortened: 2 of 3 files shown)


seabaylea:
I'm not yet convinced this is the right approach. We should be transmitting the data if MQTT is enabled (once there is a consumer for it). Once that's in place do we really need/want to have a global disablement of probe data (vs. a per-probe enable/disablement?).

I think we need to discuss the use cases and possible solutions to them before forging ahead with implementation.

tobespc:
my concern with making it reliant on MQTT is that we end up with an all or nothing approach and its tied to the transport. If we add new transports then we would have to make changes to make it work with that.  The other concern I have is that we don't at the moment have the ability to disable probes so it quite possible that users may want GC data over MQTT for example, but not the probe data.  Having it tied to the transport will not give us that flexibility.   This is also the first pass at this in the agile sense.  We can give more fine grained control in future updates

appmetrics.properties
```diff
@@ -16,3 +16,7 @@ com.ibm.diagnostics.healthcenter.logging.plugins=info
 
 # Start with method profiling enabled/disabled: on | off
 com.ibm.diagnostics.healthcenter.data.profiling=off
+
+# Allow probes to transmit data to the client: true | false
+# Defaults to false - probe data not sent.
+#com.ibm.diagnostics.healthcenter.probes.transmit=false
```
tobespc: @seabaylea  should we still be naming this com.ibm.diagnostics.healthcenter.probes.transmit?  This is purely for the appmetrics plugins for the agent so I would think "appmetrics.probes.transmit" would be a better names

hhellyeribm: I've made that change. The appmetrics plugin probably shouldn't put it's properties in the healthcenter namespace. We should avoid that to stop properties for different plugins interfering with each other.

hhellyeribm:
Closing as this isn't the implementation we are going to use initially.

PR #113, opened 2015-12-11
Enable emitting of the http data as it can be visualised.

hhellyeribm:
This is a simple change to send http metrics (only) to the attached clients. This code will be updated once data from other probes can also be visualised.

This is for issue #110


tobespc:
reviewed ok

Issue #114, opened 2015-12-13
Some socket.io data missing when io.listen() API is used


seabaylea:
When socket.io is bound to a server using the listen() API, e.g:
```'js
io = require('socket.io').listen(server);
```
no data is generated for `broadcast` events.

seabaylea:
The `listen()` API maps down to the constructor, which we're successfully instrumenting. However it looks like the `listen()` API is still pointing at the original constructor rather than the instrumented version. We therefore just need to update `listen()` to point to the instrumented constructor.

seabaylea:
Closing based on PR #115

PR #115, opened 2015-12-13
Map listen API to instrumented constructor

seabaylea:
This is for issue #114 . This remaps the listen API to point to the instrumented constructor rather than leaving it to point at the original, un-instrumented version.


tobespc:
reviewed ok

PR #122, opened 2015-12-21
API tests added

ignasbol:
Issue #123 

The test suite currently has tests for this data:
- cpu 
- memory 
- gc 
- profiling 
- environment 
- event loop

It is possible to run these tests as part of npm tests (typing in 'npm test' in appmetrics directory) or independently by running 'test_runner.js'. 

'test_runner.js' triggers 'api_tests.js' that waits for the data events to happen and has tests for API specific data, also it passes the data object to 'general_tests.js' that contains all general tests (same tests for API data and message plugin(s) data).

All tests are assertions and check if data received is sensible and as expected. Currently the tests are blocking (i.e. one failed test would cause the test suite to terminate).

There are other tests for probes that could be run independently.


seabaylea:
Hi @ignasbol. Could you raise an issue to cover the need to add tests so we can link the pull request to the issue, and also describe in the pull requests the approach that you've taken so that its possible to evaluate whether that's the right design/architecture/implementation?

tunniclm:
LGTM. With `1.0.7` going out, we can merge this now.

tunniclm:
Thanks @ignasbol for resolving the conflicts.

Issue #120, opened 2015-12-17
Keystone application cannot run with appmetrics enabled trace


xiaoshimimi:
Tested appmetrics with latest keystone application: 

https://github.com/keystonejs/keystone

with 'trace' enabled, visit http://ip:port, there will be a exception, and the page cannot load:
**Screenshot**
![defect_number_keystone1](https://cloud.githubusercontent.com/assets/5269532/11865288/1393ab16-a4de-11e5-8a53-7db390a10ebd.PNG)

**Exception:**
```
called_non_callable Error thrown for request: /
TypeError: undefined is not a function
    at exports.initLocals (/opt/node-v0.10.33/sample/keystoneJS/keystonejs/route        s/middleware.js:35:2)
    at f (/opt/node-v0.10.33/sample/keystoneJS/keystonejs/node_modules/ibm-apm/n        ode_modules/appmetrics/probes/trace-probe.js:120:26)
    at async.eachSeries.asyncFinished (/opt/node-v0.10.33/sample/keystoneJS/keys        tonejs/node_modules/keystone/node_modules/grappling-hook/index.js:239:24)
    at iterate (/opt/node-v0.10.33/sample/keystoneJS/keystonejs/node_modules/key        stone/node_modules/grappling-hook/node_modules/async/lib/async.js:146:13)
    at Object.async.eachSeries (/opt/node-v0.10.33/sample/keystoneJS/keystonejs/        node_modules/keystone/node_modules/grappling-hook/node_modules/async/lib/async.j        s:162:9)
    at iterateAsyncMiddleware (/opt/node-v0.10.33/sample/keystoneJS/keystonejs/n        ode_modules/keystone/node_modules/grappling-hook/index.js:225:8)
    at /opt/node-v0.10.33/sample/keystoneJS/keystonejs/node_modules/keystone/nod        e_modules/grappling-hook/index.js:585:5
    at dezalgofy (/opt/node-v0.10.33/sample/keystoneJS/keystonejs/node_modules/k        eystone/node_modules/grappling-hook/index.js:192:2)
    at methods.callHook (/opt/node-v0.10.33/sample/keystoneJS/keystonejs/node_mo        dules/keystone/node_modules/grappling-hook/index.js:584:4)
    at /opt/node-v0.10.33/sample/keystoneJS/keystonejs/node_modules/keystone/lib        /core/mount.js:444:12
[...]

PR #127, opened 2016-01-08
Fix for keystone undefined error

DanCunnington:
Fix for issue #120


tunniclm:
@seabaylea, @tobespc: Could one of you review this? @DanCunnington and I worked on this together so I don't want to review it myself. Thanks.

seabaylea:
Its a shame that method.length is read only, so we can't easily generate a replacement function of the correct length. This looks like a reasonable functional solution.

Merging.

PR #144, opened 2016-02-04
Allow CPU profiling to be emitted in v8 profiling format

aaroncollins:
Added functionality to change the format of CPU profiling data between V8 profiling and the format appmetrics usually produces.

Profiling data produced when in v8 mode will be emitted in 60 second intervals while regular data is made available every 5 seconds.


hhellyeribm:
Merging, as with the others on this branch we will do a full review before merging into master.

PR #146, opened 2016-02-04
Fix accidental removal of objecttracker.cpp from sources list.

hhellyeribm:
npm installs but cannot run due to missing symbol.


hhellyeribm:
Previous commit accidentally removed this.

PR #148, opened 2016-02-05
Emit heap events from appmetrics-api.js

DanCunnington:


hhellyeribm:
As with the other commits, the adaptor branch will be reviewed before it is merged back into master.

PR #153, opened 2016-02-15
Change heap plugin to be on a timer

stalleyj:


hhellyeribm:
We will do a full code review before merging adapter back into master.

PR #154, opened 2016-02-16
Add a probe for the axon npm module.

hhellyer:


hhellyer:
We will do a full review when we merge the adapter branch back into master.

PR #155, opened 2016-02-16
Strong mq probe

hhellyer:
Add probe support for the strong-mq npm module.


hhellyer:
We will do a full code review when we merge the adapter branch back into master.

PR #160, opened 2016-02-24
Fix streaming downloads using request

### LINKED ISSUE #159: Appmetrics 1.0.7 fails to install from github
When installing from github we see the following error during install:
```shell
Failed to gunzip: unexpected end of file
```

When trying to use appmetrics we see this error:
```shell
Failed to open library /home/tunniclm/Projects/appmetricstst/node_modules/appmetrics/libagentcore.so:/home/tunniclm/Projects/appmetricstst/node_modules/appmetrics/libagentcore.so: cannot open shared object file: No such file or directory
module.js:355
  Module._extensions[extension](this, filename);
                               ^
Error: Failed to initialize Agent Core library
    at Error (native)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/home/tunniclm/Projects/appmetricstst/node_modules/appmetrics/index.js:25:13)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
```

This problem is introduced by PR #137 for issue #136 which changed the way files are downloaded to use the `request` package instead of http. This was intended to enable downloads from behind a proxy, however, the implementation was not correct and from that point on downloads would not work.

This only affected appmetrics installs from github (and not npm) since the `download_all_binaries.js` used in the npm package was not updated by #137 and still used the old download code.

tunniclm:
We recently realised that downloads were failing on the latest version. This PR attempts to fix the failures and also ensure the change that caused the issue (enabling support for download behind a proxy) continue to work.
@DanCunnington kindly tested this works behind a proxy.

Fixes #159

### DIFF
--- a/download_binaries.js
+++ b/download_binaries.js
@@ -96,27 +96,27 @@ var getAgentCorePlatformVersionDownloadURL = function() {
 };
 
 var downloadAndExtractTGZ = function(downloadURL, destDir) {
-	/* Downloading the binaries */
-	request(downloadURL, function(error, response, body) {
-		if (error) {
-			console.log('Got an error: ' + error.message);
-	 		process.exit(1);
-		} else {
-			console.log('Downloading and extracting tgz from ' + downloadURL + ' to ' + destDir);
-			if (response.statusCode != 200) {
-				console.log('ERROR: Unable to download ' + downloadURL);
-				process.exit(1);
-			}
-			response.pipe(zlib.createGunzip()).on('error', function(e) { 
-				console.log("Failed to gunzip: " + e.message);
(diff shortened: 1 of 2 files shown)


hhellyer:
Merging, since this broke the install from github via npm install we can close the issue once we've done:
npm install github:RuntimeTools/appmetrics
just to confirm. (It has been tested on another clone already.)

PR #163, opened 2016-02-25
Fixed issue with instrumenting leveldown on multiple 'open' calls 

aaroncollins:
Previously we were instrumenting the leveldown object after each call to open, causing more events to be emitted than we needed. Fixed to only instrument the leveldown object once.

(Repeat of previous pull request #162, but with error corrected)


DanCunnington:
Fix for #201

tobespc:
LGTM

PR #164, opened 2016-03-02
Add extra null checks null out fields in MessageData

hhellyer:
This should guard against allocation failures and a possible bug where the cleanup function is called twice by the libuv code.


seabaylea:
I've reviewed this and it looks good. Please don't merge your own PRs as is bypasses an important check on appropriateness and quality.

Issue #139, opened 2016-01-29
Memory Leak in about 3~7 hours 


xiaoshimimi:
We found a problem that, with "request" enabled (without "request" enabled case is still in test), the node.js application will begin to have memory leak at about 3~7 hours from the application start. Depend on how many IO (logs printed). More IO, memory leak begin later. 

Once the application become memory leak, the memory, cpu, and gc events are all stop triggering. only http event still occur. And the return data of http event is correctly.

I will add more detailed information in mail.

tobespc:
Just as an update, we have recreated the OOM problem and are now working on fixing it.  Will update again once I have made some progress

tobespc:
for info, this issue is currently being dicussed in an email thread

xiaoshimimi:
In without "request" enabled case, the memory leak is still happened.

xiaoshimimi:
Current test result:
disable GC receiver, remove all probes, disable socket sender: Memory won't leak
remove all probes, disable socket sender: memory won't leak

xiaoshimimi:
Just test a case that only using appmetrics (with acmeair v2), and find a leak at about 23 hours. A mail is sent to talk about this.

tobespc:
looks like the fix we have under  git://github.com/stalleyj/appmetrics#apm resolves this memory leak.  We need to check that we can use the npm module we are referencing before releasing.  Our performance tests look good with the change in

PR #168, opened 2016-03-03
remove serialiser and just use JSON.stringify

tobespc:


tobespc:
rewrite of fix for issue #139

hhellyer:
Looks good, will need to do something else if we starting sending request data as that contains circular references by design and JSON.stringify() will throw errors for that. That can be done under a separate issue.

Issue #172, opened 2016-03-10
Add support for Http Status


saxis:
Would be helpful to get the http responses as part of the http monitor.

tobespc:
@saxis Yes, we can add that, I'll put a change together this week

tobespc:
@saxis Are you just after the status code or are are there any other fields you be interested in?

saxis:
Content-type, the header itself also if possible.

tobespc:
@saxis does this change meet your requirements if you now pull a build from git ?

tobespc:
@saxis have you have had a chance to run tests with the updated http response data ?   I'll add it to the next release if it meets your requirements

saxis:
Sorry for the delayed response. Yes our needs are met, thanks very much.

PR #176, opened 2016-03-16
add http response data

tobespc:
new values added to http for issue #172


hhellyer:
Change looks good and user is happy. Merging.

PR #185, opened 2016-04-05
Move calls to uv_async_init so they occur on the main thread.

hhellyer:
This patch moves all our calls to uv_async_init back onto the main thread.
That also means changing how we pass our messages to the event loop as we can no longer put them in the data field of the uv_async_t structure.
I've added a linked list to store the messages to send and some simple locking to make sure the two threads don't interfere with one another. We need to use the list as the libuv docs say that two sends of the same callback may be coalesced. See: http://docs.libuv.org/en/v1.x/async.html?highlight=uv_async_send#c.uv_async_send
In practise this just means that when we receive the event we deal with all the messages that are on the list not just the latest one.

### DIFF
--- a/src/appmetrics.cpp
+++ b/src/appmetrics.cpp
@@ -389,8 +389,7 @@ static void sendData(const std::string &sourceId, unsigned int size, void *data)
 	 */
 	void* dataCopy = malloc(size);
 	payload->source = new std::string(sourceId);
-	uv_async_t *async = new uv_async_t;
-	if ( NULL == dataCopy || NULL == payload->source || NULL == async ) {
+	if ( NULL == dataCopy || NULL == payload->source ) {
 		freePayload(payload);
 		return;
 	}
--- a/src/appmetrics.cpp
+++ b/src/appmetrics.cpp
@@ -43,6 +43,25 @@ static std::string* appmetricsDir;
 static bool running = false;
 static loaderCoreFunctions* loaderApi;
 
+struct MessageData {
+	const std::string* source;
+	void* data;
+	unsigned int size;
+	MessageData* next;
+};
+
+static MessageData* messageList = NULL;
+static uv_mutex_t _messageListMutex;
+static uv_mutex_t* messageListMutex = &_messageListMutex;
+static uv_async_t _messageAsync;
+static uv_async_t *messageAsync = &_messageAsync;
+
+struct Listener {
+	Nan::Callback *callback;
+};
+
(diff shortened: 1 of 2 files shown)


src/appmetrics.cpp
```diff
@@ -388,10 +397,22 @@ static void sendData(const std::string &sourceId, unsigned int size, void *data)
 	memcpy(dataCopy, data, size);
 	payload->data = dataCopy;
 	payload->size = size;
-
-	async->data = payload;
-	uv_async_init(uv_default_loop(), async, (uv_async_cb)emitMessage);
```
tunniclm: Looks like `async` is no longer used, but it is still declared (https://github.com/RuntimeTools/appmetrics/pull/185/files#diff-2419e406d80ff8bfd9b4fc942cc6b491R392)

src/appmetrics.cpp
```diff
@@ -388,10 +397,22 @@ static void sendData(const std::string &sourceId, unsigned int size, void *data)
...
+	payload->next = NULL;
+
+	// Put the next message on the end of the queue.
+	// (So they are sent in the same order we added them.)
+	uv_mutex_lock(messageListMutex);
+	MessageData** tail = &messageList;
+	while( *tail != NULL ) {
+		tail = &((*tail)->next);
+	}
+	(*tail) = payload;
```
tunniclm: Could we maintain a tail pointer to prevent the need to traverse the list?
Have we considered having a max limit on the size of the list? We could always drop messages if it gets too big (eg if something is hogging the event loop for a while.

hhellyer: We could, the list is just there to handle uv_async_send calls being coalesced.
If something was hogging the event loop the previous version of the code would have kept adding new uv_async_t's to the event loop. At least now we have visibility and can make that decision in our code if we need to. (I don't think a new problem has been introduced but you are right it might be an improvement we need in the future.)

src/appmetrics.cpp
hhellyer: I'll remove that, we aren't just declaring it but new'ing a new one and leaking memory. (Good spot.)

src/appmetrics.cpp
tunniclm: Yes, this is a definite improvement I would say. I wouldn't delay merging this change for either of those comments. Might be worth revisiting at some point though.

tunniclm:
LGTM.

Issue #186, opened 2016-04-07
Error while trying run npm test


seth-priya:
I am trying the following steps on ubuntu 14.04 14.04, Trusty Tahr to build and test appmetrics from source
//To build and install
git clone https://github.com/RuntimeTools/appmetrics
cd appmetrics && npm install node-gyp nan@2.x && node_modules/.bin/node-gyp configure rebuild && cd .. && npm install ./appmetrics
//To run the tests
cd appmetrics && npm test

Is the correct way to run the tests?

I am getting this error 
root@hj-ibmibm1492:~/PriyaS/appmetrics# npm test

> appmetrics@1.0.8-dev.0 test /root/PriyaS/appmetrics
> node tests/test_runner.js

[Thu Apr  7 13:11:43 2016] com.ibm.diagnostics.healthcenter.loader INFO: Node Application Metrics 1.0.8-dev.0.201604071309 (Agent Core 3.0.6.201511030958)
[Thu Apr  7 13:11:43 2016] com.ibm.diagnostics.healthcenter.mqtt INFO: Connecting to broker localhost:1883
Running API tests:
GC tests running ...
GC tests passed succesfully.
GC tests running ...
GC tests passed succesfully.
GC tests running ...
GC tests passed succesfully.
GC tests running ...
GC tests passed succesfully.
Memory tests running ...
Memory tests passed succesfully.
Node environment tests running ...

assert.js:86
  throw new assert.AssertionError({
        ^
AssertionError: Appmetrics version format not recognised1.0.8-dev.0.201604071309, expected 99.99.99.123456789012
    at Object.runNodeEnvTests (/root/PriyaS/appmetrics/tests/general_tests.js:281:5)
    at API.<anonymous> (/root/PriyaS/appmetrics/tests/api_tests.js:57:12)
    at API.emit (events.js:104:17)
    at API.formatOSEnv (/root/PriyaS/node_modules/appmetrics/appmetrics-api.js:87:41)
    at API.raiseEvent (/root/PriyaS/node_modules/appmetrics/appmetrics-api.js:36:17)
    at events (/root/PriyaS/node_modules/appmetrics/appmetrics-api.js:180:9)
npm ERR! Test failed.  See above for more details.

If I try to comment line root/PriyaS/appmetrics/tests/general_tests.js:281, tests proceed but crash after a while:
Event loop tests running ...
Event loop tests passed succesfully.
[...]

PR #187, opened 2016-04-07
Fix npm test failure on dev versions

tunniclm:
The tests assumed the build identifier would look like one from a
full release and did not recognise a build identifier from a dev
release.

For example, this build id for release version 1.0.7 was recognised:
    1.0.7.201602110606

...but this build id for development version 1.0.8-dev.0 was not:
    1.0.8-dev.0.201604071706

This commit changes the test to recognise the build identifer from
a dev release.

Related: #186


tobespc:
LGTM

PR #188, opened 2016-04-07
Fix crash on shutdown

tunniclm:
PR #185 changed some structs to be static instead of
dynamically allocated. Unfortunately the corresponding
deallocations were not removed causing some environments
to abort() when freeing the invalid (non-heap) pointers.

Fixes: #186


tobespc:
LGTM

PR #197, opened 2016-04-25
Adapter

tobespc:


tobespc:
merging http changes from master into adapter branch

Issue #190, opened 2016-04-11
Error in `node': free(): invalid pointer: 0x000001000a79f858 


seth-priya:
Hi,

I am not sure if this is in anyway related to the issue reported earlier:
https://github.com/RuntimeTools/appmetrics/issues/186

I am observing an issue with the latest master branch on Ubuntu 15.10 (both x86 and ppc64le). I have not run into this on Ubuntu 15.04 yet or on Red Hat.

Basically, the tests are aborted and exit as soon as they start

Here is what I see on the console:
test@pts00449-vm29:~/appmetrics$ npm test

> appmetrics@1.0.8-dev.0 test /home/test/appmetrics
> node tests/test_runner.js

*** Error in `node': free(): invalid pointer: 0x000001000a79f858 ***
Aborted
npm ERR! Test failed.  See above for more details.

I have tried this with node version v4.1.1 and 0.12 and the following steps:
git clone https://github.com/RuntimeTools/appmetrics
cd appmetrics && npm install node-gyp nan@2.x && node_modules/.bin/node-gyp configure rebuild && cd .. && npm install ./appmetrics
cd appmetrics && npm test

Any inputs here would be of great help. Thanks!

Priya

hhellyer:
I have seen an issue running on Ubuntu 15.10, I believe it's a linking problem with the appmetrics libraries.

sam-github:
I have the same problem, @tobespc, on master, running `npm test`:

```
> appmetrics@1.0.9-dev.0 test /home/sam/w/ibm/appmetrics
> node tests/test_runner.js

*** Error in `node': free(): invalid pointer: 0x0000000003859658 ***
Aborted (core dumped)
npm ERR! Test failed.  See above for more details.
ibm/appmetrics (fix-install-assumptions *%) % uname -a
Linux samtu 4.2.0-35-generic #40-Ubuntu SMP Tue Mar 15 22:15:45 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
ibm/appmetrics (fix-install-assumptions *%) % node --version
v4.2.1
ibm/appmetrics (fix-install-assumptions *%) % npm --version
2.14.7
ibm/appmetrics (fix-install-assumptions *%) % g++ --version
g++ (Ubuntu 5.2.1-22ubuntu2) 5.2.1 20151010

ibm/appmetrics (fix-install-assumptions *%) % cat /etc/issue
Ubuntu 15.10 \n \l

```

[...]

PR #198, opened 2016-04-25
Fix recursive require of appmetrics

sam-github:
A package such as appmetrics cannot require itself by name
(`require('appmetrics')`), its unresolvable. If and only if appmetrics happens
to be installed as a dependency of a package will that work.

An example of when it doesn't is its own unit tests. If you clone appmetrics
and run its unit tests, they fail:

	ibm/appmetrics (adapter % u=) % npm test

	> appmetrics@1.0.9-dev.0 test /home/sam/w/ibm/appmetrics
	> node tests/test_runner.js

	module.js:339
	    throw err;
	    ^

	Error: Cannot find module 'appmetrics'

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -17,7 +17,7 @@
 var timer = require("./timer.js");
 var util = require('util');
 var domain = require('domain');
-var am = require('appmetrics');
+var am = require('../');
 
 /*
  * Global Request Tracking
--- a/tests/test_runner.js
+++ b/tests/test_runner.js
@@ -23,7 +23,7 @@ process.argv.forEach(function(elem) {
 var agent;
 if (!global)
 {
-  agent = require('appmetrics');
+  agent = require('../');
   agent.start();
 
   // Make agent visible for other script files.


sam-github:
I couldn't run tests for this, blocked on https://github.com/RuntimeTools/appmetrics/issues/190

stalleyj:
I've looked into getting the tests to run by doing a git clone and then npm test and I don't think it's possible. The fundamental parts of appmetrics are in the addon which wouldn't be present without an npm install of appmetrics first.  

Is there a different method for testing addons ?   Can we just document the requirement to install appmetrics to get the tests to run ?

sam-github:
@stalleyj I installed the addon. I didn't mean literally `git clone; npm test`, that wouldn't work with any node package, I meant the standard sequence: `git clone ...; cd ..; npm install; npm test`. And this PR fixes one problem with that scenario. There may be others, but at that point I block on the SEGV.

sam-github:
As for your questions (which aren't related to the problem this PR fixes):

> Is there a different method for testing addons ?

No, npm install then test is standard. Noone should expect a project's tests (in any language) to run before the project is built/installed.

> Can we just document the requirement to install appmetrics to get the tests to run?

Probably should, though its obvious to any node developer.

tobespc:
LGTM

Issue #205, opened 2016-05-04
appmetrics transmitting garbage for "events" over mqtt socket in 1.0.10


tobespc:
The 1.0.10 release is transmitting garbage over the MQTT for the "events" data. This means we no longer display the data correctly in Health Center

PR #207, opened 2016-05-04
Fix for event data being transmitted incorrectly

aaroncollins:
Event data now showing correctly in Health Center and via the appmetrics api

Fix for issue #205


hhellyer:
Looks correct, removes the C++ types from the calls to the public interface.

PR #219, opened 2016-05-17
Rewrote and refactored tests use TAP testing

aaroncollins:
- Switched out old assert tests for TAP tests
- Updated the package.json so npm test now works
- Moved the functionality of general_tests into api_tests and deleted it
- We now run api_tests which pulls in the test runner. This can be changed in the future


tobespc:
LGTM

PR #223, opened 2016-05-19
Improved API profiling tests and modified test_runner.js

aaroncollins:
- Profiling tests have been updated to perform tests on each key of every function in a given profiling data object. When a profiling test fails it will now print out the failing value with the TAP output

- Renamed test_runner.js to test_app.js to better reflect the refactoring changes: it no longer runs any tests itself, it simply runs a loop and requires the agent.

- test_app can be ran using '-g' flag to test that agent starts up and doesn't crash the process (intended for testing with node-hc)

### DIFF
--- a/tests/api_tests.js
+++ b/tests/api_tests.js
@@ -14,14 +14,14 @@
  * limitations under the License.
  *******************************************************************************/
  
-var runner = require('./test_runner');
-var monitor = runner.agent.monitor();
-runner.agent.enable("profiling");
+var app = require('./test_app');
+var monitor = app.agent.monitor();
+app.agent.enable("profiling");
 
 var tap = require('tap');
 tap.plan(6); // NOTE: This needs to be updated when tests are added/removed
 tap.tearDown(function(){
-	runner.endRun();
+	app.endRun();
 });
 
 var completedTests = {}; //Stores which tests have been run, ensures single run per test
--- a/tests/api_tests.js
+++ b/tests/api_tests.js
@@ -348,10 +348,10 @@ function runProfilingTests(profData, t){
 		}
 	}
 	
-	testValuesAreNotNegative("self");
-	testValuesAreNotNegative("parent");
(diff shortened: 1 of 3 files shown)


tests/api_tests.js
```diff
@@ -331,89 +329,56 @@ function runNodeEnvTests(nodeEnvData, t) {
...
 		}
-		
-		
-		if (currentFunction.count < 0){
-			t.fail(prefix + "does not have a positive (or 0) sample count (" + currentFunction.count + ")");
+		t.pass("Value of '" + keyName + "' is an integer for all functions");
+	}
+	
+	//Check the same key for all functions in data aren't negative
+	function testValuesAreNotNegative(keyName){
```
hhellyer: If you change this to testValuesAreGreaterThan(keyName, val) then you can get rid of the special case for self and put the commend where you call testValuesAreGreaterThan('self', 1);

tests/test_app.js
```diff
@@ -14,22 +14,33 @@
...
-if (!global)
-{
-  agent = require('../');
-  agent.start();
 
-  // Make agent visible for other script files.
-  module.exports.agent = agent;
+//If running global test, run long enough to ensure the agent has loaded and process doesn't crash
+if (global) {
+	var duration_secs = process.argv[2] || 10; //Default 10 seconds for global tests
```
hhellyer: Is there a way to test if the agent has started up and remove the wait 10 seconds?
(Otherwise this is likely to fail randomly on slower machines!)

aaroncollins: As we expect that we're the only people going to be running these global tests (other people will be using npm test), we can set the duration to be something larger to compensate for the slower machines

tests/api_tests.js
aaroncollins: Good point, I've added this change

hhellyer:
LGTM, thanks.

PR #224, opened 2016-05-20
Merge Opensource branch into master for using omr-agentcore

tobespc:
omr-agentcore has now been opensourced so this change is to make appmetrics use this as a subcomponent


hhellyer:
Tested installing from branch, looks good.

PR #227, opened 2016-05-24
Covers node-hc failure test case using -f flag

aaroncollins:
Needed to cover the test case of starting the agent while running with node-hc which should fail with an error message of conflicting installations


tobespc:
LGTM

PR #228, opened 2016-05-24
Changed npm test TAP output and fixed test bug

aaroncollins:
I've been told that more verbose output is preferable for our TAP tests so I've changed our package.json to reflect this.

Also fixed a small bug I introduced in the last commit


tobespc:
LGTM

PR #229, opened 2016-05-26
remerge omr-agentcore to pull from master branch

tobespc:


tobespc:
merging as urgent test required

PR #230, opened 2016-06-03
Limit the heap histogram to only include heap objects of type HeapGraphNode::kObject

hhellyer:
Not everything on the v8 heap is an "object", this does depend on exactly what your definition of object is. This changes our histogram to use a much narrower definition, it now only returns counts of HeapGraphNodes from the snapshot when HeapGraphNode.GetType() == HeapGraphNode::kObject.


hhellyer:
I've updated the change to include arrays and strings based on the description in v8-profiler.h.
I used the names (string) and (array) as these are already used in the profiling views of the Chrome dev tools for displaying strings and arrays in heap dumps so this should match what users see when examining memory usage that way.

tobespc:
LGTM

PR #235, opened 2016-06-14
remove preinstall script for omr-agentcore

tobespc:


tobespc:
having git clone in pre install breaks anyone cloning appmetrics localling and then npm installing from local repo.

hhellyer:
LGTM

PR #246, opened 2016-06-24
Escape strings and double quotes in the json string fields

DanCunnington:
For adapter, in reference to @tunniclm 's line note


DanCunnington:
Also removed the define statement for the profiling interval, as this is now set depending on json_enabled

mattcolegate:
LGTM

PR #249, opened 2016-06-24
fix for lrtime node 10

DanCunnington:
In node 10, the new array, number and newFromUnsigned integer functions only accept 1 argument.


tobespc:
LGTM

DanCunnington:
For #264

PR #250, opened 2016-06-24
don't do object tracking on node 10

DanCunnington:
For adapter


DanCunnington:
Including fixes for the heap plugin and loop plugin

hhellyer:
Unnecessary version check from objecttracker.hpp has been removed, just to tidy up.
Looks good, merging.

PR #257, opened 2016-07-11
Trace integration

DanCunnington:
expose configure function in appmetrics to allow a strong-trace instance to be passed in for tracing. Probes have been modified to wrap the callback with a call to strong-trace when monitoring finishes.


tobespc:
LGTM

DanCunnington:
For #264

PR #261, opened 2016-07-25
cpu profiling fix for json mode

DanCunnington:
emit the json in profiling json mode instead of trying to parse the message string


tobespc:
lgtm

DanCunnington:
For #264

PR #262, opened 2016-07-26
refactored express probe

DanCunnington:
Changed the express probe so a user can use the monitoring API to monitor express applications, instead of configuring express with the `use` function.


tobespc:
lgtm

DanCunnington:
Fix for #265

PR #266, opened 2016-07-29
Fix for express probe

DanCunnington:


DanCunnington:
This includes null check on newTarget.application so when using express 2 the probe doesn't crash. I've also updated the readme notifying users that we have only tested with express 4

tobespc:
lgtm

DanCunnington:
Fixup to coincide with issue #265

Issue #271, opened 2016-08-03
tests don't run from a git clone


sam-github:
```
% npm test

> appmetrics@1.0.14-dev test /home/sam/w/sn/appmetrics
> tap --reporter tap tests/api_tests.js

TAP version 13
    # Subtest: tests/api_tests.js
module.js:442
    throw err;
    ^

Error: Cannot find module './appmetrics'
    at Function.Module._resolveFilename (module.js:440:15)
    at Function.Module._load (module.js:388:25)
    at Module.require (module.js:468:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/home/sam/w/sn/appmetrics/index.js:25:13)
    at Module._compile (module.js:541:32)
    at Object.Module._extensions..js (module.js:550:10)
    at Module.load (module.js:458:32)
    at tryModuleLoad (module.js:417:12)
    at Function.Module._load (module.js:409:3)
    1..0
```

What am I doing wrong? Any setup steps should be in the package scripts, or at least mentioned in CONTRIBUTING.md.

sam-github:
CONTRIBUTING should also mention that cloning with submodules is mandatory

sam-github:
I did a completely fresh checkout, and it all worked. :perplexed:

PR #272, opened 2016-08-03
Move tap to dev dependency, not runtime

sam-github:
Don't know if the tests pass, because of https://github.com/RuntimeTools/appmetrics/issues/271, but I was noticing that appmetrics had a an unusual number of dependencies when installed as a dependency of strong-supervisor.


tobespc:
LGTM

PR #277, opened 2016-08-05
Run the tests from `npm test`

sam-github:
Why don't all the tests get run by `npm test`?


sam-github:
@hhellyer @tobespc ?

sam-github:
the tests all pass, btw

sjanuary:
@sam-github I think we don't run the probe tests by default because we don't necessarily have a test environment set with e.g. a MongoDB database every time we want to do a test run.

Issue #273, opened 2016-08-04
license field in package is invalid, and license state is unclear


sam-github:
https://github.com/RuntimeTools/appmetrics/blob/master/package.json#L28 is not a valid license.

It could be just `Apache-2.0`, or it could be `"license": "SEE LICENSE IN NOTICES.html", but there is no way defined by npm's version of SPDX to assert a combination of licenses, unless every combined license is OSI approved.

As far as I can tell, the license is just Apache-2.0, but if IBM's lawyers feel they would like to express the license only as a license file, then the second form is the correct form.

sam-github:
Fixed by d90af2262ac4f3d46

PR #281, opened 2016-08-17
nan and node-gyp to devDependencies

DanCunnington:
Only needed when installing from GitHub.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,6 @@
     "install": "node-gyp rebuild"
   },
   "directories": {
-    "licenses": "licenses",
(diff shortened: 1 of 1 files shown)


DanCunnington:
Also fixed #273

tobespc:
LGTM

sjanuary:
@tobespc I don't think this is right, I think we use NAN in the plugins e.g. https://github.com/RuntimeTools/appmetrics/blob/master/src/plugins/node/env/nodeenvplugin.cpp#L29

PR #287, opened 2016-09-08
Do not export lrtime if it is not implemented

sam-github:
If lrtime is not implemented (and it is currently only implemented on
Linux), then do not export it as a function, allowing users of
appmetrics to implement a simple feature test to see if it is available.

@sjanuary @stalleyj This blocks strong-supervisor on non-linux platforms. The previous implementation of `appmetrics.lrtime()` would throw with "undefined is not a function" on non-linux platforms.

### DIFF
--- a/index.js
+++ b/index.js
@@ -241,9 +241,7 @@ module.exports.monitor = function() {
     return this.api;
 };
 
-module.exports.lrtime = function() {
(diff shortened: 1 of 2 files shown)


sjanuary:
LGTM

PR #288, opened 2016-09-13
Let npm handle calling of node-gyp

sam-github:
node-gyp support is built into npm, using it as a direct dependency just
makes installs slower, and the package more complex.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,12 +13,10 @@
     "tar"
   ],
   "devDependencies": {
-    "node-gyp": "3.x",
     "tap": "7.x"
   },
   "scripts": {
-    "test": "tap --reporter tap tests/api_tests.js",
-    "install": "node-gyp rebuild"
+    "test": "tap --reporter tap tests/api_tests.js"
   },
   "directories": {
     "lib": "lib",


sjanuary:
This works for me, I cloned your branch and then did npm install and npm test in the appmetrics directory to test.  I'm not sure why we had node-gyp specified as a dependency in the first place though.  Does anyone else know?  @hhellyer @DanCunnington?

sam-github:
@sjanuary rebased

package.json
```diff
@@ -13,12 +13,10 @@
     "tar"
   ],
   "devDependencies": {
-    "node-gyp": "3.x",
     "tap": "7.x"
   },
   "scripts": {
-    "test": "tap --reporter tap tests/api_tests.js",
-    "install": "node-gyp rebuild"
```
sjanuary: At the moment our build scripts rely on this "install" line being here and do a simple search and replace when building the npm package. I'd prefer to leave this line in, I know node-gyp rebuild is the default but I don't think it causes any problems to be explicit and it would save us having to rework the build scripts to do something more complicated.

sjanuary: Can we not leave this line in and still remove the dev dependency?

sam-github: no, node-gyp isn't in your PATH. npm has a version internally that it uses, but its not accessible to you. by specing it as a dev dep, it gets installed as a dep, and ./node_modules/.bin/ is added to the PATH by npm when running scripts so that dependency's scripts are in the path. Its ok, this is just a minor cleanup, if it causes cascading changes in the build tooling, its not worth it (at least now).

PR #280, opened 2016-08-15
Add CPU watchdog feature

sjanuary:
adapted from Strong Agent

PR #292, opened 2016-09-15
Use Locals instead of Handles

sjanuary:
Handle has been deprecated in Node and will at some point be removed.  Node add-ons should use Local instead of Handle when implementing the initialize method.  This PR also fixes a build error on Node 0.10 since #280, as watchdog.h ported from strong-agent was already using Locals.


hhellyer:
This looks good, Handle is deprecated so needs to be removed at some point. Fixing a build break seems like a good time.

PR #293, opened 2016-09-15
Fix cpu watchdog build error on Node 0.10

sjanuary:


hhellyer:
LGTM - Change should only affect Node 0.10

Issue #275, opened 2016-08-04
upgrade to node-gyp@3.x


sam-github:
I'd PR a change for this, but I'm not sure how to test.

Despite that it jumps a major, I suspect that it does not introduce a compatibility problem for appmetrics (and anyhow, the use of graceful-fs@3 that node-gyp@2 pulls in is going to break real soon know in node).

The change that caused the node-gyp major to bump is https://github.com/nodejs/node-gyp/pull/708, FYI, which is why I don't think its going to be a problem.

sam-github:
Actually, there is no need for appmetrics to depend on node-gyp, its builtin to node. I suggest deleting both these lines:

- https://github.com/RuntimeTools/appmetrics/blob/master/package.json#L23
- https://github.com/RuntimeTools/appmetrics/blob/master/package.json#L18

sjanuary:
I think we've got a separate PR for removing the node-gyp dependency so I'm going to close this now

NikCanvin:
separate PR for removing the node-gyp dependency   https://github.com/RuntimeTools/appmetrics/pull/288

Issue #256, opened 2016-07-06
nsp check 


majgis:
```
(+) 2 vulnerabilities found
┌───────────────┬─────────────────────────────────────────────────────────────────┐
│               │ Regular Expression Denial of Service                            │
├───────────────┼─────────────────────────────────────────────────────────────────┤
│ Name          │ minimatch                                                       │
├───────────────┼─────────────────────────────────────────────────────────────────┤
│ Installed     │ 2.0.10                                                          │
├───────────────┼─────────────────────────────────────────────────────────────────┤
│ Vulnerable    │ <=3.0.1                                                         │
├───────────────┼─────────────────────────────────────────────────────────────────┤
│ Patched       │ >=3.0.2                                                         │
├───────────────┼─────────────────────────────────────────────────────────────────┤
│ Path          │ myApp@1.2.0 > appmetrics@1.0.12 > tap@5.7.3 > nyc@… │
├───────────────┼─────────────────────────────────────────────────────────────────┤
│ More Info     │ https://nodesecurity.io/advisories/118                          │
└───────────────┴─────────────────────────────────────────────────────────────────┘
┌───────────────┬─────────────────────────────────────────────────────────────────┐
│               │ Regular Expression Denial of Service                            │
├───────────────┼─────────────────────────────────────────────────────────────────┤
│ Name          │ hawk                                                            │
├───────────────┼─────────────────────────────────────────────────────────────────┤
│ Installed     │ 1.1.1                                                           │
├───────────────┼─────────────────────────────────────────────────────────────────┤
│ Vulnerable    │ < 3.1.3  || >= 4.0.0 <4.1.1                                     │
├───────────────┼─────────────────────────────────────────────────────────────────┤
│ Patched       │ >=3.1.3 < 4.0.0 || >=4.1.1                                      │
├───────────────┼─────────────────────────────────────────────────────────────────┤
│ Path          │ myApp@1.2.0 > appmetrics@1.0.12 > tap@5.7.3 > code… │
├───────────────┼─────────────────────────────────────────────────────────────────┤
│ More Info     │ https://nodesecurity.io/advisories/77                           │
└───────────────┴─────────────────────────────────────────────────────────────────┘```

busybeaver:
I guess updating the `tap` dependency would fix this...

tobespc:
@majgis Can this be closed now that that tap dependancy has been updated ?

majgis:
I ran nsp check and found no issues.  Thank you!

PR #295, opened 2016-09-21
Update dependencies for node-gyp and tap and reinstate request dependency

sjanuary:
I have run this change through the appmetrics build system and the tests are passing on all Node levels and all platforms so it should be safe to update these dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,8 +13,8 @@
     "tar"
   ],
   "devDependencies": {
-    "node-gyp": "2.x",
-    "tap": "^5.7.x"
+    "node-gyp": "3.x",
+    "tap": "7.x"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


sjanuary:
fixes #275 and fixes #256

sjanuary:
Also need to reinstate request dependency, which is required by the npm package we publish

sam-github:
@sjanuary you don't need request. It was left in `extract_all_binaries.js` by mistake, its unused. The require can be deleted, then the bundling can be deleted, and I'm pretty sure its not used by anything else in appmetrics (see `git grep "require.*request"`), so it can be deleted globally.

sjanuary:
@sam-github You're right - I had seen the build failure and assumed the dependency was removed in error as we had had another dependency removed in error at the same time.

PR #297, opened 2016-09-26
Update binding.gyp for CPU watchdog new source files

sjanuary:


sjanuary:
Actually don't think this is needed, I'm going to close this PR

PR #302, opened 2016-10-10
Re-remove request dependency - not needed

sjanuary:
'request' dependency was removed, but this caused a build breakage in the appmetrics build system so it was reinstated.  Turns out it's definitely not needed and can be safely removed.


tobespc:
LGTM

Issue #300, opened 2016-10-07
appmetrics removes its own files during install


sam-github:
The intention was probably to save disk space, but the side effect is that once appmetrics is installed, it can't be reinstalled: `npm rebuild` fails.

This is made worth because npm2 under some conditions appears to install modules *twice* even though they only appear once on disk. We see this in CI on Windows boxes mostly. This wouldn't be such a big issue, except the second time the install script runs it fails, the install tgzs were all deleted after the first install.

rmg:
I'm not able to find the troublesome install scripts in this repo to attempt a fix.. am I missing something?

sam-github:
The script isn't in this repo, the actual packaging is closely tied to the
multi-platform build system, which aggregates the binaries from all the
platforms. I think the script should be moved here, though.
​

sjanuary:
https://github.com/RuntimeTools/appmetrics/pull/303

tobespc:
Closing as fixed in #300

sjanuary:
Build system is now using this version, changes will go into 1.1.1

PR #303, opened 2016-10-11
Add extract_all_binaries.js

sjanuary:
This file is used as part of the build process, and replaces "node-gyp rebuild" in package.json when installing appmetrics from npm.


sjanuary:
fixes #300

tobespc:
LGTM

PR #306, opened 2016-10-12
Don't force MQTT broker to on

### LINKED ISSUE #209: disable mqtt
This log message appears despite configuration:
```
[Wed May  4 09:58:44 2016] com.ibm.diagnostics.healthcenter.mqtt INFO: Connecting to broker localhost:1883
```

Configuration:
```
com.ibm.diagnostics.healthcenter.mqtt=off
```

Am I wrong in assuming that this configuration should result in no attempt to connect to the broker?  Is there another configuration I should be using if I only want to use the API?

### LINKED ISSUE #305: Should be possible to switch off connecting to MQTT broker
Not every user will want appmetrics to automatically connect to an MQTT broker on the default port and start sending data. E.g. they may be running an MQTT broker for something else. It should be possible to disable this.

sjanuary:
fixes #305 and fixes #209

### DIFF
--- a/src/appmetrics.cpp
+++ b/src/appmetrics.cpp
@@ -292,8 +292,6 @@ NAN_METHOD(start) {
 
         loaderApi->init();
 
-        // Force MQTT on for now
-        loaderApi->setProperty("com.ibm.diagnostics.healthcenter.mqtt", "on");
         loaderApi->start();
     }
     if (!initMonitorApi()) {


tobespc:
LGTM

sam-github:
Does it still default to "on"?

tobespc:
Yes, still defaults to on

sjanuary:
See https://github.com/RuntimeTools/appmetrics/blob/master/appmetrics.properties#L2

Issue #301, opened 2016-10-10
trace-probe cause "Cannot convert undefined or null to object" error


xiaoshimimi:
APM Node.js got the following defects, which indicate appmetrics' trace-probe.js (in attach function) should have the same defect:

https://developer.ibm.com/answers/questions/308277/cant-use-the-monitoring-and-analytics-diagnostics.html?utm_campaign=answers&utm_medium=email&utm_source=answers-new-question&utm_content=answers-answer-question

to fix this defect, I change 
```if(Object.keys(target.prototype).length==0 && Object.keys(target).length == 0){```
to
```if(target && target.prototype && Object.keys(target.prototype).length==0 && Object.keys(target).length == 0){```

PR #311, opened 2016-10-14
Fix for trace-probe.js bug

JulienArnold:
changed line 47 in trace-probe.js to fix the bug raised here: https://github.com/RuntimeTools/appmetrics/issues/301


tobespc:
LGTM

PR #316, opened 2016-10-19
Don't duplicate AgentExtensions.h from omr-agentcore

sjanuary:
This caused appmetrics to break completely (failing with SEGFAULT) when we updated omr-agentcore to version 3.0.10 and had mismatched versions of AgentExtensions.h.  This PR adds omr-agentcore/src as an include directory instead and removes the duplicated file from appmetrics/src.  An additional typedef in the appmetrics version is added to appmetrics.cpp instead.


mattcolegate:
Sian assures me this built cleanly, I've reviewed the code changes and they make sense, and I've checked that all instances of #include AgentExtensions.h have been altered to use the more relative path by this PR. So LGTM!

Issue #320, opened 2016-11-07
Make nodereport and heapdump available via Appmetrics API call


mattcolegate:
NodeReport: https://github.com/nodejs/nodereport
heapdump: https://github.com/bnoordhuis/node-heapdump

Functions to expose:
require('nodereport').triggerReport()
require('heapdump').writeSnapshot()

mattcolegate:
Will need to include these packages in dependencies and buildDependencies, then make them available from appmetrics. I'll reuse the function names for clarity.

sam-github:
I think we need more functions for nodereport, we also need to configure the conditions under which it auto-dumps (uncaught exception, signal, etc.)

How about just exposing `appmetrics.nodereport`, so we can call anything we want on nodereport? And perhaps the same with heapdump?

mattcolegate:
I'm confused as to what value we're adding now - it seems like we're just adding the word 'appmetrics' to the front of things. What is the benefit to the end user of this work? Surely they can just require nodereport and heapdump themselves?

sam-github:
@mattcolegate the value is in having them pre-compiled and distributed as binaries, see https://ci.nodejs.org/job/node-test-pull-request/4806/

Compiling native addons requires:

1. network access to download the node headers (only for the first build, but still a blocker on customer sites)
2. a correctly installed compiler toolchain (python version correct, visual studio version correct, a c++11 compiler - not present on some LTS corporate linux distros, etc.)

The latter is particularly onerous for nodereport... things are already going badly for a customer, so we ask them to use nodereport, and now they need to a compiler set, possibly in violation of their devops policies on production machines, possibly just difficult on Windows.

mattcolegate:
So this has become a larger piece of work very quickly - as well as the basic functionality being exposed via an Appmetrics API call, there is the issue of compiled code vs separate binaries, as well as getting the calls to work via the GUI. So in order to get a delivery in, this issue will be simplified as follows:

**This issue** will deal solely with requiring and exposing the nodereport and heapdump functionality via an Appmetrics API call. Both packages will be required from npm and therefore compiled from source.

**Another issue** will be raised to discuss provision of nodereport and heapdump via source code compilation, pre-built binaries on npm, absorbsion by appmetrics (I'll edit this later with the actual issue when it is raised)

**Yet another issue** will be raised when getting the GUI work done - the actual GUI work will be done in a work item, but we will need a PR for making the node env plugin a receiver so that it can talk to the GUI.

PR #321, opened 2016-11-07
Add nodereport and heapdump calls via appmetrics API

mattcolegate:


mattcolegate:
#321

mattcolegate:
and the correct issue tag is #320

sam-github:
@mattcolegate you can edit comments in github if you want to correct them (unlike RTC)

index.js
```diff
@@ -265,3 +265,14 @@ module.exports.setJSONProfilingMode = function(val) {
...
 }
+
+module.exports.triggerReport = function() {
+    return require('nodereport').triggerReport();
+}
+
+module.exports.writeSnapshot = function() {
+    return require('heapdump').writeSnapshot.apply(null, arguments);
+}
+
```
sam-github: trailing whitespace

package.json
```diff
@@ -7,10 +7,14 @@
   },
   "dependencies": {
     "nan": "2.x",
-    "tar": "2.x"
+    "tar": "2.x",
+    "nodereport": "1.x",
```
sam-github: you added these by hand? if you used npm, they would be sorted alphabetically, and if the next person who adds a dep uses npm, it will sort them all, creating `git diff` churn, so its better to follow npm convention and keep them sorted.

sam-github:
This will help supervisor, thanks!

README.md
```diff
@@ -208,6 +208,12 @@ Allows custom monitoring events to be added into the Node Application Metrics ag
 * `type` (String) the name you wish to use for the data. A subsequent event of that type will be raised, allowing callbacks to be registered for it.
 * `data` (Object) the data to be made available with the event. The object must not contain circular references, and by convention should contain a `time` value representing the milliseconds when the event occurred.
 
+### appmetrics.writeSnapshot([filename],[callback])
```
sam-github: I would call this "writeHeapSnapshot", as-is, its not possible to guess the purpose of the API from the name - it could be a snapshot of many things.

binding.gyp
```diff
@@ -57,7 +57,7 @@
 
   "targets": [
     {
-      "target_name": "agentcore",
+      "target_name": "omr-agentcore",
```
mattcolegate: Agreed - it's needed to build from source locally on Windows. I'll move them to a new pull request.

package.json
```diff
@@ -7,10 +7,14 @@
...
+    "heapdump": "0.x"
   },
   "bundleDependencies": [
-    "tar"
+    "tar",
+    "nodereport",
+    "heapdump"
   ],
```
sjanuary: These don't need to be bundled, we only bundle tar because it's used by the install script (extract_all_binaries.js) and npm doesn't guarantee that deps will be installed before the install script is run.

package.json
mattcolegate: Alphabetised

index.js
```diff
@@ -20,7 +20,7 @@ var os = require("os")
 //var serializer = require('./lib/serializer.js');
 var aspect = require('./lib/aspect.js');
 var request = require('./lib/request.js');
-var fs = require('fs');
+var fs = require('fs');
```
mattcolegate: removed

README.md
mattcolegate: Agreed

README.md
```diff
@@ -208,6 +208,12 @@ Allows custom monitoring events to be added into the Node Application Metrics ag
...
+Dumps the v8 heap via `heapdump`. For more information, see [the heapdump README](https://github.com/bnoordhuis/node-heapdump/blob/master/README.md)
+
+### appmetrics.triggerReport()
```
mattcolegate: Yep, ok

package.json
mattcolegate: Removed

sam-github:
This needs a squash, maybe it goes without saying.

index.js
```diff
@@ -265,3 +265,11 @@ module.exports.setJSONProfilingMode = function(val) {
...
     return jsonProfilingMode;
 }
+
+module.exports.triggerNodeReport = function() {
+    return require('nodereport').triggerReport();
+}
+
+module.exports.writeHeapSnapshot = function() {
+    return require('heapdump').writeSnapshot.apply(null, arguments);
+}
```
sam-github: You are missing a new-line on the last line of the file.

README.md
```diff
@@ -208,6 +208,12 @@ Allows custom monitoring events to be added into the Node Application Metrics ag
 * `type` (String) the name you wish to use for the data. A subsequent event of that type will be raised, allowing callbacks to be registered for it.
 * `data` (Object) the data to be made available with the event. The object must not contain circular references, and by convention should contain a `time` value representing the milliseconds when the event occurred.
 
+### appmetrics.writeHeapSnapshot([filename],[callback])
+Dumps the v8 heap via `heapdump`. For more information, see [the heapdump README](https://github.com/bnoordhuis/node-heapdump/blob/master/README.md)
+
+### appmetrics.triggerNodeReport()
```
sam-github: I think it would be better to export `appmetrics.nodereport = require('nodereport')`, so we have the full nodereport API available. The advantage of having these modules in appmetrics is that they are pre-compiled (which is huge), not that the API is wrapped. Particularly with nodereport which runtimes and the community are building together and designing the API. heapdump is a bit different, it only has one method, and it hasn't changed in forever.

tobespc: @sam-github yep, I agree, especially as there is going to be additional api's on nodereport that we will want to call so that we can inject additional information

index.js
mattcolegate: Added

README.md
sam-github: Not a good point. Are you saying nodereport is a singleton? That only one instance of it can be required at any time, and it will somehow break if required multiple times, or if multiple instances of it are required into the same application? Thats not a well behaved npm package, and should be fixed in nodereport.

Requiring nodereport should have zero-side effects, and does (to my knowledge), so I don't see why requiring it from appmetrics should be an issue. It would be possible to lazy-require it by defining `appmetrics.nodereport` as a property getter.

sjanuary:
NodeReport doesn't support Node 0.12, we're still building 0.12 binaries for appmetrics so this needs to be considered.  Actually we were still building 0.10 and 5 in the last release but I guess we should drop those anyway...

sam-github:
- 0.12, I suggest that `appmetrics.nodereport === undefined` on 0.12, makes it easy to check for, I can't see any value in working to make nodereport work on something that is end of support in less than 2 months.

- 0.10, should stop  supporting in next release, but ***PLEASE*** bump the appmetrics major to 2.x, its causing all kind of problems for use when entire node versions are dropped by upstream packages in minor, or even patch, publishes. :-(. Also, please update the engines field to `>-0.12` when you drop 0.10 support.

- 5.x, hm, I guess it gets no maintenance or support from node community anymore? I guess it should be dropped too, then. Same caveats about bumping the major apply (the major can be bumped once to 2.x if both 0.10 and 5.x get dropped in the same release). Also, the engine field might need to become a comma seperated list, at that point, I'm not so familiar with it, but its becoming more useful and important now that s many node versions are in play, some of them incompatible, and yarn also enforces engines by default (unlike npm, which warns by default).

package.json
```diff
@@ -6,7 +6,9 @@
     "node-hc": "bin/appmetrics-cli.js"
   },
   "dependencies": {
+    "heapdump": "0.x",
     "nan": "2.x",
+    "nodereport": "1.x",
```
sam-github: this would have to be moved to `optionalDependencies`, because it doesn't build on 0.12 which is supported by appmetrics.

Also, both heapdump and nodereport should be listed below in bundleDependencies, so they get bundled.

And also, I belive that appmetrics actually uses/generates a slightly different package.json when building the appmetrics package for release, one that doesn't depend on nan, for example, so @sjanuary, that one will need updating to reflect the bundle spec changes.

sjanuary: It's binding.gyp that gets replaced entirely, package.json just has one part replaced inline ("node-gyp rebuild" to "node extract_all_binaries.js") so I think we'll be ok with this change.

sam-github:
This should probably be 2 PRs.

Thee is an email going around, but I'll comment here: I think heapdump should just have its source ripped out and incorporated into appmetrics. Its allowed by license, and I asked @bnoordhuis last time I talked to him, and he was fine with that.

Its not much code, it makes it a first class appmetrics feature, one that will be accessible via mqtt to the new health centre UI. The code is really stable, barely changed in the last year or so, should be low maintenance burden.

nodereport is a bit different, its an active community effort. In my mind, adding `appmetrics.nodereport` is a tactical thing, it gets a pre-built binary of nodereport on npmjs quickly, and allows strong-supervisor to incorporate it easily, so it can become a feature of apiconnect (on collectives and docker). The strategic thing would be to modify the nodereport build process on the community jenkins to build it multi-platform (which would also be useable easily by strong-supervisor).

mattcolegate:
Closing this pull request to refactor - the work has become too big to fit in 1 PR!

Issue #333, opened 2016-11-16
TypeError: Cannot read property 'filename' of undefined


richardlau:
node-hc on the master branch is failing with TypeError if run without specifying a file to run with Node v6.9.1.
```
-bash-4.2$ npm install -g runtimetools/appmetrics
...
-bash-4.2$ node-hc
/dev/shm/usenode.riclau/node-v6.9.1-linux-x64/lib/node_modules/appmetrics/index.js:17
var main_filename = require.main.filename;
                                ^

TypeError: Cannot read property 'filename' of undefined
    at Object.<anonymous> (/dev/shm/usenode.riclau/node-v6.9.1-linux-x64/lib/node_modules/appmetrics/index.js:17:33)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/dev/shm/usenode.riclau/node-v6.9.1-linux-x64/lib/node_modules/appmetrics/launcher.js:54:13)
    at Module._compile (module.js:570:32)
-bash-4.2$
```

This worked previously, e.g. with appmetrics 1.1.2:
```
-bash-4.2$ npm install -g runtimetools/appmetrics#1.1.2
...
-bash-4.2$ node-hc
[Wed Nov 16 08:04:06 2016] com.ibm.diagnostics.healthcenter.loader INFO: Node Application Metrics 1.1.2.201611160803 (Agent Core 3.0.10)
[Wed Nov 16 08:04:06 2016] com.ibm.diagnostics.healthcenter.mqtt INFO: Connecting to broker localhost:1883
> .exit
-bash-4.2$
```

richardlau:
Likely suspect is #327. cc @tobespc

tobespc:
yeah, thats my change

PR #334, opened 2016-11-16
fix issue #333

tobespc:
add a check around require.main incase we have not actually started a node program with node-hc


stalleyj:
LGTM

PR #337, opened 2016-11-17
Implement zip function for headless mode

sjanuary:


sam-github:
You can cleanup the history (particularly that merge) with `git rebase upstream/master` (replace `upstream` with whatever your upstream is).

sjanuary:
@sam-github - we usually just do "squash and merge", so that the history on head is right and don't worry too much about the history in the PRs. Is there an advantage to doing git rebase over this?

sam-github:
If the merges don't conflict, rebase has no advantage, other than cosmetic (it confuses the commit history above), and can make at least me nervous that it will be committed as-is - I've seen merges from master merged back to master on other projects. But you won't do that, so its all good.

tobespc:
LGTM

Issue #336, opened 2016-11-16
Cannot call start a second time


stalleyj:
If you try calling start(), stop() and then start() again, it fails and the process ends.

baens:
Not just can't call start/stop/start. If you do this start/stop/start cycle, things break. Like `setInterval`.

Here is a simple log output of running start/stop/start in a JS file with the debug output set in appmetrics.

[debug.txt](https://github.com/RuntimeTools/appmetrics/files/607599/debug.txt)

baens:
Also, here is a fine demonstrating that start/stop/start cycle breaking `setInterval`

```
var appmetrics = require('appmetrics');

function start() {
  console.log('--------------------START----------------------------');
  appmetrics.start();
}

function stop() {
  console.log('---------------------STOP----------------------------');
  appmetrics.stop();
}

setInterval(() => console.log('tick'), 1500);

const run = [stop, start];

function timeout() {
  console.log('timeout');
  if (run.length == 0) return;
  const next = run.shift();
  next();
  setTimeout(timeout, 5000);
}

setTimeout(timeout, 5000);
```

stalleyj:
This will be investigated further and fixed in a future release

baens:
So, I just did a quick dive of disabling plugins, and I believe it has to to with the [Node Loop Plugin](https://github.com/RuntimeTools/appmetrics/blob/master/src/plugins/node/loop/nodeloopplugin.cpp).

What it could be no idea, but at least when I remove that from my repository it seems to work.

sjanuary:
Now fixed by #348

PR #348, opened 2016-11-23
Fix a problem with double initialization

baens:
The check_handle has a problem with double initialization. So instead of trying to initialize on start, do it when the plugin is initialized and just start and stop the listening loops.

This addresses some of #336. There is still problems with the system reconnecting after the 2nd start, but that will probably need to be addressed separately.

### DIFF
--- a/src/plugins/node/loop/nodeloopplugin.cpp
+++ b/src/plugins/node/loop/nodeloopplugin.cpp
@@ -68,13 +68,17 @@ static void GetLoopInformation(uv_timer_s *data) {
 #else
 static void GetLoopInformation(uv_timer_s *data, int status) {
 #endif
+	double mean = std::numeric_limits<double>::infinity();
+	if (num != 0) {
+		mean = sum / num;
+	}
 
 	std::stringstream contentss;
 	contentss << "NodeLoopData";
 	contentss << "," << min;
 	contentss << "," << max;
 	contentss << "," << num;
-	contentss << "," << sum / num;
+	contentss << "," << mean;
 	contentss << '\n';
 
 	std::string content = contentss.str();
--- a/src/plugins/node/loop/nodeloopplugin.cpp
+++ b/src/plugins/node/loop/nodeloopplugin.cpp
@@ -140,30 +140,31 @@ extern "C" {
 	}
 
 	NODELOOPPLUGIN_DECL int ibmras_monitoring_plugin_init(const char* properties) {
+		uv_check_init(uv_default_loop(), &check_handle);
+		uv_unref(reinterpret_cast<uv_handle_t*>(&check_handle));
+
+		plugin::timer = new uv_timer_t;
+		uv_timer_init(uv_default_loop(), plugin::timer);
+		uv_unref((uv_handle_t*) plugin::timer); // don't prevent event loop exit
+
 		return 0;
 	}
 
 	NODELOOPPLUGIN_DECL int ibmras_monitoring_plugin_start() {
 		plugin::api.logMessage(fine, "[loop_node] Starting");
 
-		uv_check_init(uv_default_loop(), &check_handle);
 		uv_check_start(&check_handle, reinterpret_cast<uv_check_cb>(OnCheck));
-		uv_unref(reinterpret_cast<uv_handle_t*>(&check_handle));
+		uv_timer_start(plugin::timer, GetLoopInformation, LOOP_INTERVAL, LOOP_INTERVAL);
 
-
-        plugin::timer = new uv_timer_t;
-		uv_timer_init(uv_default_loop(), plugin::timer);
-		uv_unref((uv_handle_t*) plugin::timer); // don't prevent event loop exit
-
-        uv_timer_start(plugin::timer, GetLoopInformation, LOOP_INTERVAL, LOOP_INTERVAL);
-
-        return 0;
+		return 0;
 	}
 
 	NODELOOPPLUGIN_DECL int ibmras_monitoring_plugin_stop() {
 		plugin::api.logMessage(fine, "[loop_node] Stopping");
-        uv_timer_stop(plugin::timer);
-		uv_close((uv_handle_t*) plugin::timer, cleanupHandle);
+
+		uv_timer_stop(plugin::timer);
+		uv_check_stop(&check_handle);
+
 		return 0;
 	}
 


src/plugins/node/loop/nodeloopplugin.cpp
```diff
@@ -136,30 +139,31 @@ extern "C" {
 	}
 
 	NODELOOPPLUGIN_DECL int ibmras_monitoring_plugin_init(const char* properties) {
-		return 0;
+       uv_check_init(uv_default_loop(), &check_handle);
```
baens: So, if anyone knows `libuv` better than I do (which honestly I don't know anything about how it works). There is probably a better way of releasing the event loops on start and stop. But this way seems to work ok.

sam-github: /to @bnoordhuis?

src/plugins/node/loop/nodeloopplugin.cpp
```diff
@@ -109,7 +112,7 @@ pushsource* createPushSource(uint32 srcid, const char* name) {
 }
 
 void OnCheck(uv_check_t* handle) {
-	const uv_loop_t* const loop = handle->loop;
+    const uv_loop_t* const loop = handle->loop;
```
bnoordhuis: Indent error.  You seem to be using spaces whereas everything else in this file uses tabs.

src/plugins/node/loop/nodeloopplugin.cpp
bnoordhuis: I know a thing or two about libuv but I don't know what "releasing the event loops" means in this context.

src/plugins/node/loop/nodeloopplugin.cpp
baens: My bad, I will get that fixed.

src/plugins/node/loop/nodeloopplugin.cpp
bnoordhuis: If I had to guess, I'd say you called `uv_check_init()` again before the `uv_close()` call completed, corrupting libuv's internal structures.  Init once and start/stop on demand is fine though; an inactive check handle consumes zero libuv resources.

src/plugins/node/loop/nodeloopplugin.cpp
baens: This should now be fixed.

baens:
This should now be ready for merging. The approach I took looks to be an OK approach by way of the conversation had with @bnoordhuis 

If there is anything else that is needed, let me know.

src/plugins/node/loop/nodeloopplugin.cpp
```diff
@@ -68,13 +68,16 @@ static void GetLoopInformation(uv_timer_s *data) {
 #else
 static void GetLoopInformation(uv_timer_s *data, int status) {
 #endif
+    double mean = std::numeric_limits<double>::infinity();
+    if(num != 0)
+        mean = sum / num;
```
bnoordhuis: This change seems unrelated?

baens: I made that change because when starting and stopping the plugin, there may be no accumulated data. So `num` might be 0 at the time this event is fired, which then causes a divide by 0. This is to protect against that corner case. 

I made that change because when starting and stopping the plugin there might be a time when the accumulated data so `num` would be 0. This is to protect against that.

bnoordhuis: Okay.  Can you split it off into a separate commit?  Rule of thumb: one change per commit and each with a commit log explaining what changed and why.

The bulk whitespace update you did makes the diff noisier than it needs to be but it's up to you to decide whether you want to leave it in or not.

Aside: I don't know if the style guide says yes or no on spaces before parens.

baens:
Completely agreed with the comment from @bnoordhuis . So I've now split those two changes into two separated isolated commits. I've hopefully removed any noise due to formatting changes.

sjanuary:
@baens we need a CLA (contributor license agreement) from you before we can accept this pull request. The current process is outlined here: https://github.com/RuntimeTools/appmetrics/blob/master/CONTRIBUTING.md.

sjanuary:
@baens We have simplified the CLA process for Node Application Metrics and now you can just confirm your acceptance by typing a statement into this pull request, e.g. "I confirm my acceptance of the project CLA for Node Application Metrics". CLA is here for reference: https://github.com/RuntimeTools/appmetrics/raw/master/NodeApplicationMetrics-V1-CLA.odt. If you can do that then we will be able to merge this in.

baens:
I confirm my acceptance of the project CLA for Node Application Metrics

~Erik Lindblom.

src/plugins/node/loop/nodeloopplugin.cpp
```diff
@@ -68,13 +68,17 @@ static void GetLoopInformation(uv_timer_s *data) {
 #else
 static void GetLoopInformation(uv_timer_s *data, int status) {
 #endif
+	double mean = std::numeric_limits<double>::infinity();
```
sjanuary: I think this should be 0 rather than infinity.  I know technically dividing a number by 0 equals infinity but in terms of calculating mean values using infinity is not right.  It will also look really weird if we try to put this data in a graph.

src/plugins/node/loop/nodeloopplugin.cpp
```diff
@@ -140,30 +140,31 @@ extern "C" {
...
 
 	NODELOOPPLUGIN_DECL int ibmras_monitoring_plugin_init(const char* properties) {
+		uv_check_init(uv_default_loop(), &check_handle);
+		uv_unref(reinterpret_cast<uv_handle_t*>(&check_handle));
+
+		plugin::timer = new uv_timer_t;
+		uv_timer_init(uv_default_loop(), plugin::timer);
+		uv_unref((uv_handle_t*) plugin::timer); // don't prevent event loop exit
+
 		return 0;
```
sjanuary: @sam-github I noticed you reviewed this PR, does this look ok to you?  I think we intentionally designed appmetrics not to do any work on init so that there's no overhead if it's not used, I'm wondering if this breaks that requirement?

sam-github: It unrefs the handles, and doesn't start them, so I don't think it should have any negative effects.

sjanuary: Ok, thanks. I'm going to accept this PR in that case.

sjanuary:
I'm going to accept this PR and then change the default mean of infinity to 0 afterwards as in https://github.com/RuntimeTools/appmetrics/pull/348#discussion_r97959058.

PR #349, opened 2016-11-24
Don't block Node shutdown

sjanuary:
Bug in headless implementation was blocking node from exiting with appmetrics running

### DIFF
--- a/src/headlessutils.cpp
+++ b/src/headlessutils.cpp
@@ -55,6 +55,7 @@ const char* headlessRecVersion = "1.0";
 void start() {
 	loop = uv_default_loop();
 	uv_async_init(loop, &async_zip, asyncfunc);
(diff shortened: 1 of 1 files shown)


tobespc:
LGTM

sam-github:
Should be easy to unit test.

sjanuary:
@sam-github - how would you write that test?  This was caught by our system tests, but they're run from a shell script.

sam-github:
I think something like test/test-no-refs-leaked.js:

```
require('../').start();
```

is basically the test, and maybe you don't need `.start()`? You'd add the test to package.json's `test` property.

You'd want to make it be valid tap, though, maybe add a `tap.assert(true);` so that at least one assertion shows up.

You'd want to make this

sam-github:
Actually, looks like the bug may manifest only on calls to start, so you might need that.

PR #350, opened 2016-11-24
Call stop() on process exit

sjanuary:
Also make appmetrics start consistent whether monitor() or start() is called.


mattcolegate:
LGTM

PR #353, opened 2016-11-28
Fix headless test failure since zip on exit

sjanuary:
Headless zip was trying to read a folder that had already been deleted by the test.  Add more robust checks for folder existence and re-order test.


stalleyj:
LGTM

PR #354, opened 2016-11-28
Allow appmetrics to load AIX .a shared libraries

mattcolegate:
This sets the right library extension for AIX shared libraries.
Requires https://github.com/RuntimeTools/omr-agentcore/pull/27


tobespc:
LGTM

PR #358, opened 2016-11-30
Incorporate node-heapdump into appmetrics

sjanuary:


sam-github:
Intention is this be used as `require("appmetrics/heapdump")`?

sjanuary:
@sam-github It can be used like that, but I think the main intention is that it will be built-in for appmetrics to use, the fact that it can be used standalone is more of a side effect.

sam-github:
@sjanuary OK, so it won't be advertised to appmetrics users?

I want to use it in strong-supervisor, it will allow us to remove a dependency on heapdump, which is a compiled module. That's OK, right?

sjanuary:
That's fine, it should work exactly the same as heapdump at the 0.3.7 level, unit tests run fine  We may be out of date for a bit if any updates go into heapdump, but it's a pretty mature codebase so I don't think we need to worry too much about that.

mattcolegate:
LGTM

PR #366, opened 2017-01-19
Node-hc should return same rc as child process

### LINKED ISSUE #363: node-hc returns 0 exit code on failure
The `node-hc` script (a.k.a. [bin/appmetrics-cli.js](https://github.com/RuntimeTools/appmetrics/blob/master/bin/appmetrics-cli.js)) returns 0 even if an error was thrown, ideally it would pass back node's return code on exit.

```bash
echo 'throw "SomeError"' >test.js
node-hc test.js
echo $?                                                 # Returns 0
node test.js
echo $?                                                 # Returns 1
```

This is also the case if appmetrics wasn't installed properly (i.e. if appmetrics failed to initialise because a dependency didn't install properly).

sjanuary:
fixes  #363
also need to add a test for this

### DIFF
--- a/bin/appmetrics-cli.js
+++ b/bin/appmetrics-cli.js
@@ -49,3 +49,7 @@ if (process.env.NODE_PATH) {
 }
 
 var p = child_process.fork(launcher_filename, args, {"execArgv": node_args})
+
+p.on('close', (code) => {
+  process.exit(code)
+})
--- a/bin/appmetrics-cli.js
+++ b/bin/appmetrics-cli.js
@@ -50,6 +50,10 @@ if (process.env.NODE_PATH) {
 
 var p = child_process.fork(launcher_filename, args, {"execArgv": node_args})
 
-p.on('close', (code) => {
-  process.exit(code)
+p.on('close', (code, signal) => {
+  if(signal) {
+    process.kill(process.pid, signal)
+  } else {
+    process.exit(code)
+  }
 })


bin/appmetrics-cli.js
```diff
@@ -49,3 +49,7 @@ if (process.env.NODE_PATH) {
 }
 
 var p = child_process.fork(launcher_filename, args, {"execArgv": node_args})
+
+p.on('close', (code) => {
```
sam-github: This will work if the child process exits, but if it terminates with signal, the cli will exit with `0`.

```
on('close', (code, signal) => {
  if(signal) {
    process.kill(process.pid, signal);
  } else {
    process.exit(code);
  }
});
```
might work better

sjanuary: Thanks @sam-github, I'll make that change.

sjanuary: PR updated

gibfahn:
FYI if you add a `Fixes: https://github.com/RuntimeTools/appmetrics/issues/363` line to your commit message, when you land this it will automatically close the issue.

sjanuary:
@gibfahn - "Fixes" with a capital F?  I used "fixes #363", does that not work?

tobespc:
LGTM

sjanuary:
@tobespc - I think you closed this instead of merging by accident

tobespc:
lgtm

gibfahn:
@sjanuary did you ever get round to adding a test for this?

PR #368, opened 2017-01-23
correction for the description of event `loop`

a8775:
Current version of the appmetrics returns the following object in the event `loop`:

``` typescript
interface AppMetricsLoop {
    count: number; // (Number) the number of event loop ticks in the last interval.
    minimum: number; // (Number) the shortest (i.e. fastest) tick in milliseconds.
    maximum: number; // (Number) the longest (slowest) tick in milliseconds.
    average: number; // (Number) the average tick time in milliseconds.
}
```


sjanuary:
@a8775 thanks for the PR, merging now.

PR #372, opened 2017-01-31
Add heapdump trigger

tobespc:


mattcolegate:
LGTM

sam-github:
This lacked a test even though its easy to write, and it had a syntax bug: #376

sjanuary:
A test for this could be based on https://github.com/RuntimeTools/appmetrics/blob/master/tests/heapdump/test-callback.js which came from the heapdump project.

PR #377, opened 2017-02-01
streamline CLA process

NikCanvin:
replaced CLA odt file, and related  text in Contributing.md


sjanuary:
Thanks for your work on this Nik, this will make it much easier for us to accept contributions to appmetrics.

PR #378, opened 2017-02-03
make start() chainable

sam-github:


index.js
```diff
@@ -130,6 +130,8 @@ aspect.after(module.__proto__, 'require', data, function(obj, methodName, args,
         for (var i = 0; i < probes.length; i++) {
             if (probes[i].name === args[0]) {
                 ret = probes[i].attach(args[0], ret, module.exports);
+                // XXX(sam) can't have multiple probes with same name, why not
+                // return from the loop immediately?
             }
```
sjanuary: @sam-github we can and do have multiple probes with the same name (http and http-outbound are both named "http" as both attach to the http module.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "appmetrics",
-  "version": "2.0.0",
+  "version": "2.1.0",
```
sjanuary: @sam-github we haven't released 2.0.0 yet...

sam-github:
@sjanuary PTAL

sjanuary:
Now merged

PR #389, opened 2017-02-16
Add travis integration

tobespc:
This add travis ci integration.  The build will fail as the we need to remove the test so that will follow.

Test has also been changed now so passed ok


stalleyj:
Looks good and will be very helpful

Issue #394, opened 2017-02-22
Test fails on 6.10.0


gibfahn:
The logic in [tests/api_tests.js](https://github.com/RuntimeTools/appmetrics/blob/82d7b9d750fb131f3cfbcf772d617703e36c41da/tests/api_tests.js#L326) for checking the node version is a bit of a hack, and can't handle Node `6.10.0` (it converts the first 2 digits to a float, and then compares that float to 6.5). `6.10` is therefore `6.1`.

```js
var nodeVersion = Number(process.version.match(/^v(\d+\.\d+)/)[1]);
if(nodeVersion >= 6.5) { // issue 283
  t.ok(2*parseInt(nodeEnvData['max.semi.space.size']) + parseInt(nodeEnvData['max.old.space.size']) === parseInt(nodeEnvData['heap.size.limit']),
  'Values for max.old.space.size and max.semi.space.size match heap.size.limit');
} else {
  t.ok(4*parseInt(nodeEnvData['max.semi.space.size']) + parseInt(nodeEnvData['max.old.space.size']) === parseInt(nodeEnvData['heap.size.limit']),
  'Values for max.old.space.size and max.semi.space.size match heap.size.limit');
}
```

#### Test failure

```js
$ node tests/api_tests.js

    not ok 10 - Values for max.old.space.size and max.semi.space.size match heap.size.limit
      ---
      at:
        line: 330
        column: 51
        file: tests/api_tests.js
        function: runNodeEnvTests
      stack: |
        runNodeEnvTests (tests/api_tests.js:330:51)
        tests/api_tests.js:215:5
        API.<anonymous> (tests/api_tests.js:208:7)
        API.raiseLocalEvent (appmetrics-api.js:240:7)
        Object.module.exports.emit (index.js:256:18)
        Timeout.latencyReport [as _onTimeout] (index.js:105:13)
      source: >
        t.ok(4*parseInt(nodeEnvData['max.semi.space.size']) +
        parseInt(nodeEnvData['max.old.space.size']) ===
        parseInt(nodeEnvData['heap.size.limit']),
      ...
```

gibfahn:
Fix in #395

PR #395, opened 2017-02-22
Fix node version check for 6.10.0

gibfahn:
The previous code converted the node version to a float and compared to 6.5. This
doesn't work for 6.10 (6.10 < 6.5).

Fixes: https://github.com/RuntimeTools/appmetrics/issues/394

### DIFF
--- a/tests/api_tests.js
+++ b/tests/api_tests.js
@@ -13,10 +13,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  *******************************************************************************/
- 
+'use strict'
+
 var app = require('./test_app');
 var appmetrics = app.start();
 var monitor = app.appmetrics.monitor();
+var semver = require('semver');
 app.appmetrics.enable("profiling");
 
 var tap = require('tap');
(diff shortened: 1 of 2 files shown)


gibfahn:
I'd note that now `nvm install 6` is installing `6.10.0`, all other PR travis runs will fail on 6 due to this issue.

tests/api_tests.js
```diff
@@ -322,8 +323,13 @@ function runNodeEnvTests(nodeEnvData, t) {
                                          t.ok(parseInt(nodeEnvData['max.old.space.size']) > 0,
                                               "max.old.space.size is positive");
 
-                                              var nodeVersion = Number(process.version.match(/^v(\d+\.\d+)/)[1]);
-                                              if(nodeVersion >= 6.5) { // issue 283
+                                              // converts e.g. 'v8.0.0' to [8,0.0]
```
sam-github: consider using https://www.npmjs.com/package/semver for version comparisons

gibfahn: What's the advantage to using semver (to outweigh the downside of having another devDependency)? This should work for all node versions, and it's < 10 lines of code.

sam-github: What is the downside to using a devDependency?

If you use semver, its 2 lines of code, with no comments, and one of the lines is `var semver = require('semver')`, which barely counts as a line of code.

I can read this code, if I have to, and work through the regex, and figure if its correct, but I'd rather not.

sam-github: Also, I think I've read a half dozen buggy attempts to compare versions by now, yesterday was https://github.com/ibmdb/node-ibm_db/pull/228#discussion_r102289716. I've looked at yours, and I think it is correct, but it really would be better if people just use `semver` unless there is a really compelling reason not to (like the fact that package dependencies cannot be relied on to be installed before an install script is run in npm@2, but that doesn't apply here).

gibfahn: Fair enough, updated, PTAL.

Looking at it more, it's a pretty tiny dep anyway, so nbd to import.

mattcolegate:
LGTM2

PR #403, opened 2017-03-01
Make appmetrics a global singleton

### LINKED ISSUE #402: appmetrics module is singleton, causing failures on multiple use
This can happen when it is used by strong-supervisor, and then used as well (perhaps indirectly as a sub-dependency) by the application being run by the supervisor.

I recommend an approach like https://github.com/strongloop/strong-cluster-control/blob/master/index.js#L12-L21

cluster-control restricts to exact version match, though matching only on major.minor may be a better strategy.

sjanuary:
Fixes #402 
@sam-github please review

### DIFF
--- a/index.js
+++ b/index.js
@@ -27,6 +27,8 @@ var fs = require('fs');
 var agent = require("./appmetrics")
 var headlessZip = require("./headless_zip.js")
 var heapdump = require('./heapdump.js');
+var VERSION = require('./package.json').version
+var assert = require('assert')
 
 
 // Set the plugin search path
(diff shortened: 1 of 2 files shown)


sjanuary:
CI failure looks unrelated to my PR, a new Node v7 has been released (7.7.0) and appmetrics doesn't build on it.

sjanuary:
Looks like 7.7.0 build failure is due to https://github.com/nodejs/node/issues/11628

tobespc:
accepting as failing build is on the latest node7 which we don't officially support yet

Issue #390, opened 2017-02-17
'request' event data contains duplicated inbound http request


xiaoshimimi:
We have an application monitored by appmetrics 2.0.1, the test request is as following:

User use: /rest/api/getAllPets to request data, application send a http://localhost:4001/api/pets request to query and return the result to user.

As the request event data cannot be stringified. I write logs to print its content iterate.
********** request event data ********** is the event data itself. 
data.request is included in ********** request event data ********** section, so we don't print it separately.
********** children of request event data ********** is each child in children and iterate.

we can see, request id=4 and request id=5 both have type=HTTP. But they are all means to /rest/api/getAllPets. They are duplicated. And actually, request id=5 don't have much meaningful context like request id=4. 
request id=6 is correct, because it is the HTTP Outbound request, which call http://localhost:4001/api/pets. 

```
[2017-02-14 01:21:06.939] [ERROR] knj_log - ********** request event data **********
[2017-02-14 01:21:06.940] [ERROR] knj_log - { time: 1487056866913,
  type: 'HTTP',
  name: '/rest/api/getAllPets',
  duration: 24.483617,
  request:
   Request {
     children: [ [Object] ],
     name: '/rest/api/getAllPets',
     type: 'HTTP',
     parent: undefined,
     id: 4,
     top: [Circular],
     timer:
      Timer {
        startTime: [Object],
        startTimeMillis: 1487056866913,
        timeDelta: 24.483617,
        cpuTimeDelta: -1 },
     active: false,
     tracedStart: true,
     context: { url: '/rest/api/getAllPets' },
     traceStopped: true,
     stack: 'Error\nat ServerResponse.res.end(/root/JanDemo/http_app/node_modules/connect/lib/middleware/logger.js:149)\nat ServerResponse.module.exports [as send](/root/JanDemo/http_app/node_modules/response-send/lib/response-send.js:108)\nat IncomingMessage.<anonymous>(/root/JanDemo/http_app/index.js:74)\nat <module>.emitOne(events.js:96)\nat IncomingMessage.emit(events.js:188)\nat IncomingMessage.Readable.read(_stream_readable.js:381)\nat <module>.flow(_stream_readable.js:761)\nat <module>.resume_(_stream_readable.js:743)\nat <module>._combinedTickCallback(internal/process/next_tick.js:74)' } }
     
[2017-02-14 01:21:06.948] [ERROR] knj_log - ********** children of request event data **********
[...]

Issue #268, opened 2016-07-29
express probe is not providing data when express.Router() method is used


tobespc:
acmeair uses express.Router() and the express probe needs to cater for this to work

jhondavila:
the same thing happens , some correcion for this problem.
try a hook but I did not work well

bhashah:
Hi,

If I use 'app.get(path, callback)' method for creating route paths, the express probe gets properly created and works just fine. But, in our application code, we are using 'app.route(path).all(callback).get(callback)' method to setup our express route paths. Appmetrics is not able to place express probes on routes created this way. Have you provided any solution yet to setup probes for app.route ?

tobespc:
closing as we have removed the express probe #390 #404

PR #260, opened 2016-07-22
Remove express probe and add strong-express-metrics probe

DanCunnington:
For strongloop parity, tested with adapter - all express tests pass and also independently using express. As this is for strong-supervisor only, no documentation has been added to the readme

PR #404, opened 2017-03-06
Change Express Request probe type to 'express'

mattcolegate:
@seabaylea - This looks to fix the issue encountered in #390 but would like you to further clarify architechture with Response type field - should they be just 'HTTP' or 'DB' or should they be related to the actual probe e.g. 'express','mqlight'? Currently in the source we seem to have a half-and-half situation.

** Please do not pull until reviewed **

### DIFF
--- a/probes/express-probe.js
+++ b/probes/express-probe.js
@@ -93,7 +93,7 @@ ExpressProbe.prototype.metricsEnd = function(probeData, methodName, methodArgs) 
 
 // Heavyweight request probes for express queries 
 ExpressProbe.prototype.requestStart = function (probeData, method, methodArgs) {
-  probeData.req = request.startRequest( 'HTTP', 'request', false, probeData.timer );
+  probeData.req = request.startRequest( 'express', 'request', false, probeData.timer );
   probeData.req.setContext({url: methodArgs[0]});
 };
 


sam-github:
Just curious, but why is there an express probe? https://github.com/mattcolegate/appmetrics/blob/24ba71e51d2c59e59712dcb9ecbf63929640bd82/probes/express-probe.js#L73-L78 metrics are  all generic, by probing the HTTP and HTTPS APIs in node (which appmetrics does), those metrics are already available for _all_ HTTP frameworks, including express.

sjanuary:
Appmetrics actually doesn't currently get https metrics although I don't think that's the reason we have an Express probe...  I actually thought it was a port from strong-agent, but obviously not if you don't know why we have one.  Agree that the data seems to duplicate the http data currently which is not that useful.

sam-github:
There was an express probe added briefly to appmetrics for strong-agent compatibility, but it was a misunderstanding, it was removed and replaced with a strong-express-metrics probe in #260 

The express probe was then re-added in #262 to fix #265, but #265 doesn't have any details other than a subject, so, the trail ends there.

@DanCunnington, can you provide background?

sjanuary:
Perhaps we should also disable the Express probe by default, with a view to either making it useful or removing it entirely in the future.  There's no point just duplicating the HTTP metrics like we are right now, that's potentially worsening performance for no actual gain.

sam-github:
If its not useful, maybe just delete it rather than disabling?

DanCunnington:
@mattcolegate @sam-github there may be background detailed in a work item in RTC - I remember re implementing the probe to match the strong-agent version (https://github.com/strongloop/strong-agent/blob/master/lib/probes/express.js)

sjanuary:
Strongloop's Express probe seems to record it as 'expressCalls'.  It also handles the Router use case, which we don't (#268).

PR #410, opened 2017-03-21
Build should fall back to node-gyp rebuild

sjanuary:


sjanuary:
Main build command would now be the same for installing from source as installing from npmjs.org.  When building from source it just checks for the existance of the binary package and if it's not there then it runs node-gyp rebuild.

gibfahn:
I think the general is to use `node-pre-gyp` here, in particular the [`--fallback-to-build`](https://github.com/mapbox/node-pre-gyp#options) option. IDK if this provides any benefit over using a home-grown solution, but it might be worth looking into.

extract_all_binaries.js
```diff
@@ -130,33 +130,54 @@ var getWindowsRedisFiles = function() {
...
+    }
 	}
 	
 };
 
+function tryRebuild() {
+  console.log('Falling back to node-gyp rebuild');
+  var childProcess = require('child_process')
+  var res = childProcess.spawnSync('node-gyp', ['rebuild'], {stdio:'inherit', shell:true});
+  if(res.signal) {
```
rmg: What happens when the parent process receives a signal instead? Does it properly terminate the child process on all platforms? @sam-github you might know the answer to this since I know you've done a deep dive in to this before..

package.json
```diff
@@ -21,7 +21,7 @@
   },
   "scripts": {
     "test": "tap --reporter tap --timeout=120 tests/*tests.js tests/probes/http-outbound-probe-test.js tests/probes/http-probe-test.js tests/headless_test.js",
-    "install": "node-gyp rebuild"
+    "install": "node extract_all_binaries.js"
```
rmg: Would it be portable to do `node extract_all_binaries.js || node-gyp rebuild` and _not_ modify `extract_all_binaries.js` to do the build as a fallback?

sjanuary: Yes, this is a nicer solution.  I have modified the PR.

tobespc:
LGTM

Issue #367, opened 2017-01-23
Error: Monitoring API is not initialized


a8775:
The following code under Windows 10 and nodejs 6.2.0 raise an error:
```
var appmetrics = require('appmetrics');
appmetrics.configure({
    applicationID: 'processId:' + process.pid, // Specifies a unique identifier for the mqtt connection
    mqtt: 'off', // Specifies whether the monitoring agent sends data to the mqtt broker. The default value is on
    mqttHost: 'localhost', // Specifies the host name of the mqtt broker
    mqttPort: '883', // Specifies the port number of the mqtt broker
    profiling: 'off' // Specifies whether method profiling data will be captured. The default value is off
});
appmetrics.disable('profiling');
```

[01/23/17 13:13:33] com.ibm.diagnostics.healthcenter.loader INFO: Node Application Metrics 1.2.0.201612012154 (Agent Core 3.1.0)
D:\test\node_modules\appmetrics\index.js:183
        agent.sendControlCommand("profiling_node", "off,profiling_node_subsystem");
              ^

Error: Monitoring API is not initialized
    at Error (native)
    at Object.module.exports.disable (D:\test\node_modules\appmetrics\index.js:183:15)
    at Object.<anonymous> (D:\test\bin\Test\Diagnostics.js:157:12)
    at Module._compile (module.js:541:32)
    at Object.Module._extensions..js (module.js:550:10)
    at Module.load (module.js:458:32)
    at tryModuleLoad (module.js:417:12)
    at Function.Module._load (module.js:409:3)
    at Timeout.Module.runMain [as _onTimeout] (module.js:575:10)
    at tryOnTimeout (timers.js:224:11)

tobespc:
Have you started the agent with appmetrics.start() ?

a8775:
Not started, this code is run like in example.

sjanuary:
This is probably a doc update, looks like 'disable' can't be called if the agent hasn't been started.  'configure' or the appmetrics.properties file can be used instead (also profiling is disabled by default for this specific example).

a8775:
Yes, you are definitely right.

```
var appmetrics = require('appmetrics');
var monitoring = appmetrics.monitor();
```

From user perspective of the API, please take look that all cited methods are based on `appmetrics` and not on ` monitoring`. Some are working and some are not - this is not very intuitive. 

Possibly it's possible to make a change in code without changing the API that unify
[...]

PR #412, opened 2017-03-22
Improve README for requests and enable/disable and order probes alphabetically

### LINKED ISSUE #405: README is not great for 'request' events
E.g. it says there are only two types, "HTTP" and "DB", which is not actually true.

Also see #392 for discussion on 'context' field being undocumented

sjanuary:
Fixes #405


sjanuary:
Fixes #367

stalleyj:
LGTM

PR #418, opened 2017-03-29
On process exit, remove latency methods from loop

mattcolegate:
This was causing an assertion error on Node 7.7.4 on Windows:

Assertion failed: ((handle)->flags & UV__HANDLE_CLOSING) == 0, file src\win\loop-watcher.c, line 121

Caused by leaving work on the loop whilst attempting shutdown.


mattcolegate:
LGTM

PR #419, opened 2017-03-30
Stop node loop from using sstream << int on Windows

mattcolegate:
This works around an issue with node 7.7.4 on Windows.


mattcolegate:
LGTM

PR #423, opened 2017-04-03
Undo Matt's changes

stalleyj:
We are currently unsure for the reason for these changes and it does brake profiling data as when pushing a double to std::to_string we get it formatted to 6 decimal places and this brakes the parsing.


tobespc:
LGTM - needs more investigations

PR #426, opened 2017-04-06
Rewrite mongo probe to support latest MongoDB API

### LINKED ISSUE #181: For db queries, return the number of records found
When hitting a database, you should report on the number of records returned. I've seen a lot of applications where a user wrote a query expecting 1 result and we're actually getting N back. By seeing this reported in debugging, they were able to improve their application by correcting the query.

### LINKED ISSUE #296: MongoDB probe missing some transactions
The MongoDB Collection API provides more methods for modifying a collection than we are actually monitoring with our probe, including  deleteMany, deleteOne, insertMany, insertOne, replaceOne, updateMany, updateOne.  We are potentially missing quite a lot of information by not advising these methods.

Full API here: http://mongodb.github.io/node-mongodb-native/2.2/api/Collection.html

sjanuary:
Fixes #296 
Fixes #181 
Also adds a test, but needs a MongoDB database to be running for the tests to pass so I have not added the test to the npm test call.

### DIFF
--- a/probes/mongo-probe.js
+++ b/probes/mongo-probe.js
@@ -41,8 +41,33 @@ MongoProbe.prototype.aspectCollectionMethod = function(coll, method) {
                     if (typeof(callbackPosition) != 'undefined') {
                         aspect.strongTraceTransactionLink('mongodb: ', methodName, methodArgs[callbackPosition]);
                     }
-
-                    that.metricsProbeEnd(probeData, collectionName, method, methodArgs);
+                    var count = undefined;
+                     
+                    if (args && args.length > 1) {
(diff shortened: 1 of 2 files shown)


sam-github:
I suggest pinging the mongo port with a tcp connect, and skip the test (`tap.test('x', {skip: 'no mongodb'}, ...`) if the DB  isn't present, so the tests are always there and run if possible.

sjanuary:
That does risk writing random test material to a Mongo database set up for other purposes though, or getting test failures because there's already a populated database.

stalleyj:
LGTM

PR #427, opened 2017-04-10
Work around stringstream and uint32_t issue on Node 7

sjanuary:
Convert uint32_t to int32_t in nodeloopplugin.cpp to work around the problem we have seen on Windows 7 with Node.js 7.7.3 onwards where pushing any uint32_t into a stringstream causes a crash.

Also made a related change to not emit data from the event loop plugin if none has been collected so that we're never emitting the default values of 9999, 0, 0 and 0 which could skew averages and graphs.

### DIFF
--- a/src/plugins/node/loop/nodeloopplugin.cpp
+++ b/src/plugins/node/loop/nodeloopplugin.cpp
@@ -56,11 +56,10 @@ static void cleanupHandle(uv_handle_t *handle) {
 }
 
 uv_check_t check_handle;
-uint32_t statistics[4];
-uint32_t& min = statistics[0];
-uint32_t& max = statistics[1];
-uint32_t& num = statistics[2];
-uint32_t& sum = statistics[3];
+int32_t min = 9999;
+int32_t max = 0;
+int32_t num = 0;
+int32_t sum = 0;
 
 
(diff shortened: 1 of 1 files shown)


tobespc:
LGTM

src/plugins/node/loop/nodeloopplugin.cpp
```diff
@@ -56,47 +56,46 @@ static void cleanupHandle(uv_handle_t *handle) {
...
-	}
 
-	std::stringstream contentss;
-	contentss << "NodeLoopData";
-	contentss << "," << min;
-	contentss << "," << max;
-	contentss << "," << num;
-	contentss << "," << mean;
-	contentss << '\n';
+	  std::stringstream contentss;
```
sam-github: why use streams here, isn't this just a really ineffecient `snprintf()`?

sjanuary: What would be more efficient?

sam-github: You could remove all memory allocation by using a buffer on the stack, snprintfing into it with a single funtion call. it would also be one line of code, so easier to read, the multi-line iostream code seems pretty cumbersome to me in comparison.

PR #431, opened 2017-04-11
Update latest-release branch to 3.0.1

mattcolegate:


sjanuary:
This includes #430, so one commit too many although we need to go one further now anyway.  Can you update the PR for 3.0.1?

tobespc:
@mattcolegate is this even needed still now that we are at 3.0.2 ?

mattcolegate:
@sjanuary I've now updated the PR to 3.0.1 commit - changed PR name to reflect this.
@tobespc Part of the release instructions I was looking at included changing the `latest-release` branch to point to the latest release (makes sense). Either we keep the `latest-release` branch pointing to the latest release or there really isn't any point having a `latest-release` branch that I can see.

mattcolegate:
After discussion with @tobespc , agreed to remove the `latest-release` branch.

Issue #433, opened 2017-04-11
http-outbound-probe-test failure with Node 8


gibfahn:
Test failure with Node 8

#### Failure:

```
[Tue Apr 11 11:10:13 2017] com.ibm.diagnostics.healthcenter.mqtt INFO: Connecting to broker localhost:1883
    1..3
    # Subtest: HTTP Outbound Event
        ok 1 - Timestamp is an integer
        ok 2 - Should report GET as HTTP request method
        ok 3 - Should report http://localhost:8000/ as URL
        ok 4 - Should report world as value of hello header
        1..4
    ok 1 - HTTP Outbound Event # time=3.623ms

    not ok 2 - timeout!
      ---
      signal: SIGTERM
      handles:
        - type: Server
          events:
            - request
            - connection
          connectionKey: '6::::8000'
      ...
    not ok 3 - missing test
    # failed 2 of 3 tests
    # time=119892.613ms
not ok 3 - tests/probes/http-outbound-probe-test.js # time=120109.473ms
```

#### To reproduce:

Change the first URL for a different platform:

```bash
mkdir -p node8
# Use this line for Mac:
curl https://nodejs.org/download/nightly/v8.0.0-nightly20170410b470a85f07/node-v8.0.0-nightly20170410b470a85f07-darwin-x64.tar.gz | tar -xz --strip-components=1 -C "$PWD"/node8/
# For Linux use this line instead:
# curl https://nodejs.org/download/nightly/v8.0.0-nightly20170410b470a85f07/node-v8.0.0-nightly20170410b470a85f07-linux-x64.tar.gz | tar -xz --strip-components=1 -C "$PWD"/node8/
curl https://nodejs.org/download/nightly/v8.0.0-nightly20170410b470a85f07/node-v8.0.0-nightly20170410b470a85f07-headers.tar.gz > node8-headers.tar.gz
export PATH="$PWD/node8/bin:$PATH"
git clone --recursive https://github.com/RuntimeTools/appmetrics.git && cd appmetrics
git checkout 3.0.1
npm i --tarball=../node8-headers.tar.gz
npm test
```

stalleyj:
I can now recreate this on the latest nightly build, continuing to investigate why our probe isn't working

stalleyj:
I don't yet know why but this problem is due to the change in the way the module exports are done

sam-github:
I think I see the problem.

```http.js
exports.get = function(...) {
   return exports.request(...);
}
[...]

PR #437, opened 2017-04-25
Fix probe for node 8 for issue #433

stalleyj:
Because of changes to the http module in node 8, we now need to probe both the request and the get method in our http-outbound probe.  Use semver to check the version of node as we must only do this on node 8, otherwise we will get double metrics on older node versions. Fixes issue #433.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
   "dependencies": {
     "nan": "2.x",
     "tar": "2.x",
+    "semver": "^5.3.0",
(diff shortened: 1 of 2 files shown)


sjanuary:
LGTM

mattcolegate:
@grams this is why semver moved from dev-dependency to dependency.

mattcolegate:
Apologies grams, I meant to tag @gdams ^^

PR #440, opened 2017-05-02
Include description of loopback events in README

### LINKED ISSUE #200: Support loopback monitoring

sjanuary:
Also renames loopback request event for consistency

Fixes #200

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,6 +324,13 @@ Emitted when a LevelDB query is made using the `leveldown` module.
     * `value` (Object) The value being added to the LevelDB database using the `put` method (Undefined for other methods) 
     * `opCount` (Number) The number of operations carried out by a `batch` method (Undefined for other methods) 
     * `duration` (Number) the time taken for the LevelDB query to be responded to in ms.
+    
+### Event: 'loopback-datasource-juggler'
+Emitted when a function is called on the `loopback-datasource-juggler` module
+* `data` (Object) the data from the loopback-datasource-juggler event:
+    * `time` (Number) the time in milliseconds when the event occurred. This can be converted to a Date using `new Date(data.time)`
+    * `method` (String) the function the juggler has executed
+    * `duration` (Number) the time taken for the operation to complete.
 
 ### Event: 'memcached'
 Emitted when a data is stored, retrieved or modified in Memcached using the `memcached` module.
--- a/probes/loopback-probe.js
+++ b/probes/loopback-probe.js
@@ -83,7 +83,7 @@ loopbackDJProbe.prototype.metricsEnd = function(probeData, method, methodArgs) {
  * Heavyweight request probes for juggler commands
  */
 loopbackDJProbe.prototype.requestStart = function (probeData, target, method, methodArgs) {
-	 probeData.req = request.startRequest( 'loopback', "query" );
+	 probeData.req = request.startRequest( 'loopback-datasource-juggler', "query" );
 	 probeData.req.setContext({loopbackDJProbe: methodArgs[0]});
 };
 


tobespc:
LGTM

README.md
```diff
@@ -324,6 +324,13 @@ Emitted when a LevelDB query is made using the `leveldown` module.
     * `value` (Object) The value being added to the LevelDB database using the `put` method (Undefined for other methods) 
     * `opCount` (Number) The number of operations carried out by a `batch` method (Undefined for other methods) 
     * `duration` (Number) the time taken for the LevelDB query to be responded to in ms.
+    
+### Event: 'loopback-datasource-juggler'
+Emitted when a function is called on the `loopback-datasource-juggler` module
```
sam-github: /cc @bajtos @raymondfeng @ritch FYI

bajtos: What kind of functions is this intercepting? Any function defined in `loopback-datasource-juggler`? Or only static & prototype methods on `Model` classes? I don't mind either, it's just something that's not clear to me from this description.

sjanuary: https://github.com/sjanuary/appmetrics/blob/5da4bafe2564dadd33b39f8937fd866356fc5ea0/probes/loopback-probe.js#L56-L57 shows the functions being intercepted and it looks like they're on `DataAccessObject`. Ideally the person who wrote the probe should have done the docs at the same time because he's not around any more and I'm trying to work out from the code what it does... I'm happy to make changes though, e.g. could list the functions in the description if that would be helpful.

sam-github: At the moment its pretty unclear what it does, it would be great if you could reverse engineer and document it.

bajtos: I see, it makes more sense now.

I am proposing to reword this sentence as follows:

> Emitted when a `DataAccessObject` function is called. `DataAccessObject` functions are mixed into models that are attached to a database-backed connector, e.g. `MySQL` or `MongoDB`. These functions are exposed to REST clients via LoopBack's `PersistedModel`, you can learn more about them in [LoopBack documentation](https://loopback.io/doc/en/lb3/PersistedModel-REST-API.html).

I am not sure if intercepting a white-list of methods is the best way forward. For example, we have recently added [`upsertWithWhere`](https://github.com/strongloop/loopback-datasource-juggler/blob/b34b6e0ad41e9e2935488ff95c13bf93d892d212/lib/dao.js#L706) that's not included in that list.

IIRC, all methods defined on `DataAccessObject` are mixed into the target model as public API methods, I think it should be safe to intercept them all by simply iterating over all constructor/prototype fields and filtering function values only.

While we are improving this probe, it would be good to intercept `KeyValueAccessObject` too, see https://github.com/strongloop/loopback-datasource-juggler/blob/b34b6e0ad41e9e2935488ff95c13bf93d892d212/lib/kvao/index.js

These two access-objects should cover most paths where LoopBack talks to different databases.

However, there are connectors that don't use any built-in access object: SOAP, REST, Swagger, possibly few more. These connectors build a custom access object depending on the API exposed by the backend service they are connecting to. Installing an intercepter for these methods may be more difficult. However, if you can identify small tweaks we can make in LoopBack to make that task easier, then I think we should include them in the framework.

cc @raymondfeng @kjdelisle

sjanuary: @bajtos I'm happy to make that README update, or you can raise a pull request if you prefer?  For the probe improvements you would also be welcome to raise a pull request, or if not then could you open a new issue to track the requirement? Thanks

sam-github:
@bajtos  Fwiw, at collab summit we discussed building a guide for how to write npm modules so that they are amenable to APM obervation, one way is for modules to expose an API that they maintain that marks events that should be observable. http://mongodb.github.io/node-mongodb-native/2.0/reference/management/apm/ is one example of such an API. LoopBack may benefit from such an API, by uniformly implementing instrumentability, it connectors could be treated more uniformly by appmetrics and others, and LB connectors that use common code would have the instrumentation for free, and ones that, like you mention, have more customized implementations would have the burden (should not be heavy) of maintaining instrumentation hooks, instead of appmetrics having to track the implementation with monkey patches. I've a backlog item to look at this, I'll make a proposal, where should it go?  Is there a Lb.next proposal repo? cc: @mhdawson

kjdelisle:
@sam-github https://github.com/strongloop/loopback-next

Love the idea :+1:

bajtos:
@sam-github I like your idea of making LoopBack easier to observe by APM too 👍 

Considering that loopback.next is still a work in progress and our plan for the initial version is to use existing juggler and connector implementations, I think we should add APM support into the current 3.x release line. In which case it makes more sense to me to discuss @sam-github's proposal in https://github.com/strongloop/loopback or even in https://github.com/strongloop/loopback-datasource-juggler

cc @raymondfeng @ritch

README.md
bajtos: @sjanuary Please raise the pull request updating README yourself if you don't mind.

> For the probe improvements you would also be welcome to raise a pull request, or if not then could you open a new issue to track the requirement? 

Makes sense. Looks like Sam has already a backlog item for this work, let's leave it at that for now.

PR #441, opened 2017-05-02
Add version check for Node 8 binaries

mattcolegate:


extract_all_binaries.js
```diff
@@ -112,6 +112,9 @@ var getSupportedNodeVersionOrExit = function() {
 	if (process.version.indexOf('v7') === 0) {
 		return '7';
 	}
+	if (process.version.indexOf('v8') === 0) {
```
gibfahn: You could also just do:

```js
if (process.version.startsWith('v8')) {
```

mattcolegate: Is that more performant?

gibfahn: >Is that more performant?

Probably not, just looks clearer.

Even nicer alternative:

```js
const supportedVersions = ['0.10','0.12','2','4','5','6','7','8'];
for (var version of supportedVersions) 
  if (process.version.startsWith(`v${version}.`))
    return version;
```

but it's mostly just cosmetic.

tobespc:
lgtm

PR #447, opened 2017-05-18
Add additional info to the emitter in order to expose it to the api. …

kgriesh:
…Adding baseUrl and route info  allows for collecting data on urls that contain path variables. I accept and agree to be bound by the terms of the IBM Contributor License Agreement.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,9 +12,6 @@
     "semver": "^5.3.0",
     "jszip": "2.5.x"
   },
-  "bundleDependencies": [
-    "tar"
-  ],
   "devDependencies": {
     "codecov": "2.x",
     "node-gyp": "3.x",
--- a/probes/http-outbound-probe.js
+++ b/probes/http-outbound-probe.js
@@ -140,8 +140,8 @@ function formatURL(httpOptions) {
     }
     if(httpOptions.host) {
        url += httpOptions.host
-    } else  if(httpOptions.hostname) {
-       url += httpOptions.host
+    } else if(httpOptions.hostname) {
+       url += httpOptions.hostname
     } else {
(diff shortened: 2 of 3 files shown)


sjanuary:
The `bundleDependencies` section in package.json is needed by a downstream dependency

package.json
```diff
@@ -12,9 +12,6 @@
     "semver": "^5.3.0",
     "jszip": "2.5.x"
   },
-  "bundleDependencies": [
-    "tar"
```
sam-github: tar is needed to run the install script. npm doesn't guarantee that a packages deps are installed before its install script is run, so the only safe way for an install script to use any deps is to have them be bundled. Whether any particular install sees this or not depends on race conditions, so while not bundling the install script's deps may work most of the time, for most people, it fails for us in CI, where we do lots of runs, across lots of systems.

sjanuary:
Please remove the change to package.json and add an update to README.md to describe the extra properties on an http event.  A test for the new properties in https://github.com/RuntimeTools/appmetrics/blob/master/tests/probes/http-probe-test.js would also be nice to have.

tobespc:
@kgriesh Hi, please can you update your pull request with the comments made by @sjanuary above.  We can then accept this request.  thanks

koushikr:
@kgriesh Can you follow and resubmit a merge please?

Waiting for one.

stalleyj:
@kgriesh This extra data is express specific, please can you explain the requirement to have these individually rather than just the url which is already part of the http event ?

stalleyj:
@kgriesh Closing this PR as we haven't heard from you, please re-open if this is something you are still interested in.

PR #449, opened 2017-05-19
Fix clang warnings

gibfahn:
Not sure if this is the right way to fix these, but with this PR and https://github.com/RuntimeTools/omr-agentcore/pull/46 there are no warnings on macOS Sierra except for:

```
clang: warning: libstdc++ is deprecated; move to libc++ with a minimum deployment target of OS X 10.9 [-Wdeprecated]
```

`npm test` passes.

caveat emptor...


mattcolegate:
LGTM

PR #274, opened 2016-08-04
Add eslint support

sam-github:
I gave up on line lengths and eqeqeq support, probes in particular would need to substantially rewritten to not have run-on lines.

There are a couple dozen errors remaining and which appear to me to be genuine bugs (eslint isn't all style).

I recommend a close review while ignoring whitespace: https://github.com/RuntimeTools/appmetrics/pull/274/files?w=1

I don't know if tests still pass, see #271

PR #455, opened 2017-06-15
Add HTTPS probes and tests

### LINKED ISSUE #409: HTTPS probe
Hi
There is an HTTP probe, what runs correctly and only for HTTP 'action'.
But, how to monitor HTTPS server (not the express part), on the same basics as HTTP probe ? I didn't anything about that in the documentation, the past issues or in the code.

sjanuary:
Fixes #409


sjanuary:
Just investigating test failures on Node 8...

gibfahn:
>Just investigating test failures on Node 8...

Make sure you're not using 8.1.1 (and if you are look for https://github.com/nodejs/node/issues/13667)

sjanuary:
Test failure was just a timing issue, the test was closing the server after sending requests but before they had reached it. Now fixed.

sjanuary:
Updated to comply with eslint #274

probes/https-outbound-probe.js
```diff
@@ -0,0 +1,210 @@
...
+// Probe to instrument outbound https requests
+
+function HttpsOutboundProbe() {
+  Probe.call(this, 'https'); // match the name of the module we're instrumenting
+}
+util.inherits(HttpsOutboundProbe, Probe);
+
+HttpsOutboundProbe.prototype.attach = function(name, target) {
+  var that = this;
+  if (name == 'https') {
```
bnoordhuis: `===`?

probes/https-outbound-probe.js
```diff
@@ -0,0 +1,210 @@
...
+      target,
+      methods,
+      // Before 'https.request' function
+      function(obj, methodName, methodArgs, probeData) {
+        // Get HTTPS request method from options
+        var options = methodArgs[0];
+        var requestMethod = 'GET';
+        var urlRequested = '';
+        var headers = '';
+        if (typeof options === 'object') {
```
bnoordhuis: `if (options !== null && typeof options === 'object') {`?

probes/https-outbound-probe.js
```diff
@@ -0,0 +1,210 @@
...
+      // After 'https.request' function returns
+      function(target, methodName, methodArgs, probeData, rc) {
+        // If no callback has been used then end the metrics after returning from the method instead
+        if (aspect.findCallbackArg(methodArgs) == undefined) {
+          // Need to get request method and URL again
+          var options = methodArgs[0];
+          var requestMethod = 'GET';
+          var urlRequested = '';
+          var headers = '';
+          if (typeof options === 'object') {
```
bnoordhuis: Likewise.

probes/https-outbound-probe.js
```diff
@@ -0,0 +1,210 @@
...
+            urlRequested = formatURL(options);
+            if (options.method) {
+              requestMethod = options.method;
+            }
+            if (options.headers) {
+              headers = options.headers;
+            }
+          } else if (typeof options === 'string') {
+            urlRequested = options;
+            var parsedOptions = url.parse(options);
```
bnoordhuis: Callling `url.parse()` before and after is arguably a little wasteful, it's not cheap.

probes/https-outbound-probe.js
```diff
@@ -0,0 +1,210 @@
...
+  if (httpsOptions.port) {
+    url += ':' + httpsOptions.port;
+  }
+  if (httpsOptions.path) {
+    url += httpsOptions.path;
+  } else {
+    url += '/';
+  }
+  return url;
+}
```
bnoordhuis: How does this differ from `url.format()`?

probes/https-outbound-probe.js
```diff
@@ -0,0 +1,210 @@
...
+HttpsOutboundProbe.prototype.metricsEnd = function(probeData, method, url, res, headers) {
+  if (probeData && probeData.timer) {
+    probeData.timer.stop();
+    am.emit('https-outbound', {
+      time: probeData.timer.startTimeMillis,
+      method: method,
+      url: url,
+      duration: probeData.timer.timeDelta,
+      statusCode: res.statusCode,
+      contentType: res.headers ? res.headers['content-type'] : 'undefined',
```
bnoordhuis: What if `res.headers['content-type'] === undefined`?

probes/https-outbound-probe.js
```diff
@@ -0,0 +1,210 @@
...
+  // Do not mark as a root request
+  probeData.req = request.startRequest(reqType, url, false, probeData.timer);
+};
+
+HttpsOutboundProbe.prototype.requestEnd = function(probeData, method, url, res, headers) {
+  if (probeData && probeData.req)
+    probeData.req.stop({
+      url: url,
+      statusCode: res.statusCode,
+      contentType: res.headers ? res.headers['content-type'] : 'undefined',
```
bnoordhuis: Likewise.

probes/https-probe.js
```diff
@@ -0,0 +1,157 @@
...
+      });
+    });
+  }
+  return target;
+};
+
+/*
+ * Custom req.url parser that strips out any trailing query
+ */
+var parse = function(url) {
```
bnoordhuis: Just `function parse(url) {`?

probes/https-probe.js
```diff
@@ -0,0 +1,157 @@
...
+  return url;
+};
+
+/*
+ * Ignore requests for URLs which we've been configured via regex to ignore
+ */
+HttpsProbe.prototype.filterUrl = function(req) {
+  var resultUrl = parse(req.url);
+  var filters = this.config.filters;
+  if (filters.length == 0) return resultUrl;
```
bnoordhuis: `===`?

probes/https-probe.js
```diff
@@ -0,0 +1,157 @@
...
+HttpsProbe.prototype.requestEnd = function(probeData, method, url, res, httpsReq) {
+  if (probeData && probeData.req)
+    probeData.req.stop({
+      url: url,
+      method: method,
+      requestHeader: httpsReq.headers,
+      statusCode: res.statusCode,
+      header: res._header,
+      contentType: res.getHeader('content-type'),
+    });
```
bnoordhuis: I'd put braces around the body for legibility.

probes/https-probe.js
```diff
@@ -0,0 +1,157 @@
...
+ * Set configuration by merging passed in config with current one
+ */
+HttpsProbe.prototype.setConfig = function(newConfig) {
+  if (typeof newConfig.filters !== 'undefined') {
+    newConfig.filters.forEach(function(filter) {
+      if (typeof filter.regex === 'undefined') {
+        filter.regex = new RegExp(filter.pattern);
+      }
+    });
+  }
```
bnoordhuis: This mutates `newConfig`.  Is that something callers expect?

tests/probes/https-outbound-probe-test.js
```diff
@@ -0,0 +1,78 @@
...
+  }
+}
+
+function isInteger(n) {
+  return isNumeric(n) && n % 1 == 0;
+}
+
+function isNumeric(n) {
+  return !isNaN(parseFloat(n)) && isFinite(n);
+}
```
bnoordhuis: Seemingly unused functions.

tests/probes/https-outbound-probe-test.js
```diff
@@ -0,0 +1,78 @@
...
+var https = require('https');
+
+var tap = require('tap');
+
+tap.plan(3);
+
+tap.tearDown(function() {
+  setTimeout(function() {
+    server.close();
+  }, 1000);
```
bnoordhuis: Is there a reason for this timeout?  The next test doesn't have one.

tests/probes/https-probe-test.js
```diff
@@ -0,0 +1,132 @@
...
+  }
+}
+
+function isInteger(n) {
+  return isNumeric(n) && n % 1 == 0;
+}
+
+function isNumeric(n) {
+  return !isNaN(parseFloat(n)) && isFinite(n);
+}
```
bnoordhuis: Unused functions.

tests/test_https_server.js
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+'use strict';
+
+var https = require('https');
+var fs = require('fs');
+var path = require('path');
+
+console.log('dirname = ' + __dirname);
```
bnoordhuis: What is this used for?

probes/https-outbound-probe.js
```diff
@@ -0,0 +1,210 @@
...
+var aspect = require('../lib/aspect.js');
+var request = require('../lib/request.js');
+var util = require('util');
+var url = require('url');
+var am = require('../');
+var semver = require('semver');
+
+var methods;
+if (semver.lt(process.version, '8.0.0')) {
+  methods = ['request'];
```
bnoordhuis: `https.get()` exists in older versions.  Is there a reason for the conditional?

sjanuary: #433

tests/probes/https-outbound-probe-test.js
sjanuary: Yes, the test was crashing because the server was being closed after the http outbound call was made but before it was received by the server, so this gives the server time to receive the call.  The other test tests incoming https calls so by definition doesn't close the server until the calls have got there.

tests/test_https_server.js
sjanuary: Debug, left in by mistake

probes/https-outbound-probe.js
sjanuary: Just googled this - I didn't realise typeof null was 'object'...

probes/https-outbound-probe.js
sjanuary: then contentType gets set to undefined, which I think is ok

tests/probes/https-outbound-probe-test.js
sjanuary: used on line 47

tests/probes/https-probe-test.js
sjanuary: lines 66, 70, 82

probes/https-outbound-probe.js
sjanuary: I'll look into this. Metrics mode definitely doesn't need it but requests seem to need a name.  We might be able to put a dummy name in though.

probes/https-outbound-probe.js
sjanuary: It strips off trailing ? or # components, which we don't use.

probes/https-probe.js
sjanuary: Yes, I think this is ok, it's because there's more than one way of setting config options and this merges them (copied from existing http-probe).  
As a side note it looks like we don't have any tests for URL filtering so that would be good to do.

sjanuary:
@bnoordhuis thanks for the comments, I've fixed or replied to all of them, just one that I need to look into further.

probes/https-outbound-probe.js
bnoordhuis: I'd perhaps add an explaining comment, it's not clear from context why node 8 is special.

probes/https-outbound-probe.js
sam-github: maybe should be `: undefined`. right now, if there is no res.headers, contentType is a string `'undefined'`, if there are res.headers, but no 'content-type', then its `undefined` (not a string), seems inconsistent.

probes/https-outbound-probe.js
```diff
@@ -0,0 +1,210 @@
...
+ *******************************************************************************/
+'use strict';
+
+var Probe = require('../lib/probe.js');
+var aspect = require('../lib/aspect.js');
+var request = require('../lib/request.js');
+var util = require('util');
+var url = require('url');
+var am = require('../');
+var semver = require('semver');
```
sam-github: I prefer requires to be sorted alphabetically (by line) rather being in some random order.

sam-github: (but don't know if that's part of the appmetrics coding style)

tests/probes/https-outbound-probe-test.js
```diff
@@ -0,0 +1,78 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+'use strict';
+
+process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
+var appmetrics = (appmetrics = require('../../'));
```
sam-github: I don't understand this line, I thought maybe you were trying to set `global.appmetrics` as well as the module-scope `appmetrics`, but that isn't how it works. Typo?

tests/probes/https-outbound-probe-test.js
```diff
@@ -0,0 +1,78 @@
...
+var options = {
+  host: 'localhost',
+  port: 8000,
+  headers: {
+    hello: 'world',
+  },
+};
+
+// Request with a callback
+https.get('https://localhost:8000/', function(res) {});
```
sam-github: maybe assert the callback is called?

tests/probes/https-outbound-probe-test.js
```diff
@@ -0,0 +1,78 @@
...
+
+tap.plan(3);
+
+tap.tearDown(function() {
+  setTimeout(function() {
+    server.close();
+  }, 1000);
+});
+
+var completedTests = 0;
```
sam-github: you don't need this, if you get more (or less) than 3 https-outbound events, then there is a bug.

tests/probes/https-probe-test.js
```diff
@@ -0,0 +1,132 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ ******************************************************************************/
+'use strict';
+
+process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
+var appmetrics = (appmetrics = require('../../'));
```
sam-github: it this test is the same as http (and it seems it is), it can be factored into a utility function, and just called with `httpTest(require('https'))` (or `http`). Besides eliminating cut-n-paste, it makes it clear that the behaviour of the probes should be identical.

sam-github: You'd also have to pass the event and monitor names ('https-outbound' vs 'http-outbound'), the rest looks the same.

tests/probes/https-outbound-probe-test.js
sjanuary: Yes, this confused me too but looking at the history there was an extra ` appmetrics =` in the original commit and I think eslint-fix or prettify must have added the brackets.

probes/https-outbound-probe.js
sjanuary: ok, makes sense

probes/https-outbound-probe.js
sjanuary: Personally I like to group related requires together, e.g. all local requires together and any chained requires together.   Then alphabetical after that. This is not my ordering though (copied) so I will improve it a bit so that it looks less random.

We don't really have a coding style beyond what eslint does and some naming/case stuff. I wouldn't like to enforce something like alphabetical ordering with eslint or otherwise because ordering can change behaviour, e.g. appmetrics itself usually needs to be required first.

probes/https-probe.js
sjanuary: done

tests/probes/https-outbound-probe-test.js
sjanuary: removed

tests/probes/https-outbound-probe-test.js
sjanuary: now fixed

probes/https-outbound-probe.js
sjanuary: I rewrote this bit in both http-outbound-probe and https-outbound-probe. They're both now only calling `url.parse()` on metrics end.

tobespc:
@sjanuary Looks like a lot of codebeat issues have been introduced with this request.  Could you take a look please, thanks

sjanuary:
@tobespc I don't really rate codebeat tbh, I'm not sure why we're using it.  The 'issues' are things like 'function too long', 'similar code in 2 places', 'too many function arguments', 'block nesting too deep'.  There are a lot introduced because there's a lot of new code here and some of it is similar to existing code, but some are not fixable because of how the probes framework work and others are not fixable for other reasons.  I've looked through the codebeat report and there's nothing I really want to change - are there any specific issues that you would like to be addressed?

tobespc:
accepting pull request whilst we look at the wider issue of codebeat and its findings

anthonywebb:
Tremendous work!  Any idea when the next version will be released that includes this https stuff?

PR #465, opened 2017-08-31
fix `exports.emit is not a function`

nqvst:
seems that the module keyword was forgotten
i accept the "IBM Runtimes Tools Contributor License Agreement".

### DIFF
--- a/index.js
+++ b/index.js
@@ -124,7 +124,7 @@ if (global.Appmetrics) {
       global.Appmetrics.VERSION +
       '.\n'
   );
-  module.exports = global.Appmetrics;
+  exports = module.exports = global.Appmetrics;
 } else {
   global.Appmetrics = module.exports;
   module.exports.VERSION = VERSION;
--- a/index.js
+++ b/index.js
@@ -98,7 +98,7 @@ var latencyReport = function() {
     max: latencyData.max,
     avg: latencyData.total / latencyData.count,
   };
-  exports.emit('eventloop', { time: Date.now(), latency: latency });
+  module.exports.emit('eventloop', { time: Date.now(), latency: latency });
   latencyData.count = 0;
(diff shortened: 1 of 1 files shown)


sam-github:
Note that this can only happen when multiple appmetrics instances exist, and second one is using the previously required instance, a situation that is hard to arrange in unit tests.

https://github.com/RuntimeTools/appmetrics/blob/master/index.js#L127 should be `exports = module.exports = global.Appmetrics;`, keeping the two always identical avoids this kind of error.

nqvst:
@sam-github  I'm sorry, but I don't understand what you mean. is exports defined somewhere? I thought it was just a typo.
do you want me to change back and then do `exports = module.exports = global.Appmetrics;`

sam-github:
https://nodejs.org/api/modules.html#modules_exports

I think your change makes things more consistent, and the change I suggest would make the code less fragile.

sjanuary:
@sam-github isn't `exports` already an alias for `module.exports` in Node.js?

nqvst:
i don't think so. in my repl node 6 and 8 it gives me `ReferenceError: exports is not defined`

sjanuary:
@nqvst see [https://nodejs.org/api/modules.html#modules_exports_shortcut](https://nodejs.org/api/modules.html#modules_exports_shortcut), which is why the current implementation works.  Sam is right though, if we reassign module.exports we should also reassign exports according to that link.

nqvst:
@sjanuary ok, thanks for the info!

sam-github:
>  isn't exports already an alias for module.exports in Node.js?

They aren't true aliases in the sense if you change one the other gets somehow changed too. But they do start out with the same value assigned to both, and are both user-assignable, so if you assign to one and not the other then they no longer have the same value.

PR #466, opened 2017-09-04
Fix ambiguous reference to 'debug' with Node 8.4.0

sjanuary:
Appmetrics doesn't currently build against the latest Node 8, due to an ambiguous reference to 'debug'.  We are trying to use `debug` from the `loggingLevel` enum in omr-agentcore which unfortunately isn't in a namespace, but prefixing with `::` resolves to the global namespace and fixes the build error.


sjanuary:
duplicate #463

PR #467, opened 2017-09-05
epoll blocked time should not be counted in loop

sam-github:
https://github.com/RuntimeTools/appmetrics/blob/4fde82994326ebce1a328b4f188d2a8a3cfce72e/src/plugins/node/loop/nodeloopplugin.cpp#L110-L125

Above code isn't doing what I would have expected.

1. its truncating loop times to integral milliseconds - that's what got me looking at it, trying to figure out why times were always integral for `loop`, but had sub-millisecond precision for `eventloop`
2. check runs right after poll returns, so its measuring https://github.com/libuv/libuv/blob/564677d325db6122cba2091100d07ea6d366161b/src/unix/core.c#L352-L363, which includes the time spent blocked waiting for io/timeout, which is not so useful

AFAICT, appmetrics should be calling uv_hrtime itself in a check callback, and then in the next prepare callback calculating the diff, which would be measuring the actual time that the loop is executing, as opposed to time blocked on epoll (time blocked on epoll is a sign of an unloaded application, not a performance problem).

/cc @bnoordhuis


sam-github:
I don't think `eventloop` is likely to every find performance problems, it would have to somehow run a `setImmediate()` in the exact same tick that a long-running or synchronous API call is made in order to show anything.

Below shows the output from a badly behaved node application that is periodically busy-looping in js, you can see the `loop` max times show that there is a problem, the `eventloop` times don't show anything.

I'd consider deprecating the `eventloop` metrics, and removing them from the dashboard.

screenshot from 2017-09-05 13-18-01

```
'use strict';

require('../').monitor({
  port: 'PORT' in process.env ? process.env.PORT : 3000,
  host: 'localhost',
  appmetrics: require('appmetrics'),
});

setInterval(function() {
  if (Math.random() < 0.9) return;

  var start = Date.now();
  while((Date.now() - start) < (Math.random() * 100))
    ;
}, 1000);

var waiting = 0;

function want() {
  // randomly choose how much load
  return Math.random() * (3 * parseInt(process.env.UV_THREADPOOL_SIZE || 4));
}

function load() {
  var w = want();
  while (waiting < w) {
    waiting++;
    require('crypto').pbkdf2('pass', 'salt', 102400, 20, 'sha512', function() {
      waiting--;
      setImmediate(load);
    });
  }
}

load();
```

sjanuary:
I don't see a problem with changing the type of metrics we record for the event loop, especially if the current latency measure is not very useful, but I think it needs some discussion with the wider team and would also need to be a major version change. Presumably removing it entirely would supersede this PR, it looks like this was originally to make the latency measure more accurate?

@seabaylea I think the event loop latency requirement came from you, any thoughts? Or was it a Strongloop parity item?

PR #468, opened 2017-09-06
Add data to the http/s event

stalleyj:
This pull request takes the changes we wanted from #447 without the package update and also adds the information to the README.md


stalleyj:
After discussion with the team we have decided to close this as there isn't a clear reason for why the additional information is needed. We have heard nothing from the original raiser.

PR #474, opened 2017-09-18
Support Node for z/OS

mattcolegate:
Please squash and merge

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@ install.log
 libagentcore.so
 licenses
 node_modules
-plugins
 .project
 /agentcore.dll
 /heapdump.node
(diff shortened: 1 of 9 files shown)


stalleyj:
@mattcolegate we need to skip the headless tests as these won't pass on z/OS at the moment and make sure headless fails gracefully on z/OS.

mattcolegate:
I think what's happened here is that I've straight commented-out headless functionality in `index.js` and `appmetrics.cpp` - will need to surround those statements with a check for ZOS / os390.
I'll also put that check into the headless tests.

mattcolegate:
@stalleyj Travis builds now pass, ready for re-review

stalleyj:
Travis is passing so let's get this in and built on the other platforms.

PR #477, opened 2017-09-27
Expose new GC types introduced in v8 version 4.6

sam-github:
Everything that was not kGCTypeMarkSweepCompact used to be mapped to "S"
(Scavenge), but that became wrong with 4.6, which has more GC types.

I've examined appmetrics-dash, it seems to treat all GC types identically, so this might not break it.

Conservatively, I would call it semver-major, but then again, maybe it can be considered a bug fix since the new GC types were introduced in node 5.x, but were all being classified as "Scavenge".

### DIFF
--- a/src/plugins/node/gc/nodegcplugin.cpp
+++ b/src/plugins/node/gc/nodegcplugin.cpp
@@ -149,7 +149,17 @@ void afterGC(v8::Isolate *isolate, GCType type, GCCallbackFlags flags) {
 	gcRealEnd = GetRealTime();
 
 	// GC type
-	const char *gcType = (type == kGCTypeMarkSweepCompact) ? "M" : "S";
+	const char *gcType = NULL;
+        switch (type) {
+          case kGCTypeMarkSweepCompact: gcType = "M"; break;
+          case kGCTypeScavenge: gcType = "S"; break;
+#if NODE_VERSION_AT_LEAST(5, 0, 0)
+          case kGCTypeIncrementalMarking: gcType = "I"; break;
+          case kGCTypeProcessWeakCallbacks: gcType = "W"; break;
+#endif
+          // Should never happen, but call it minor if type is unrecognized.
+          default: gcType = "S"; break;
+        }
 
 	// GC heap stats
 	HeapStatistics hs;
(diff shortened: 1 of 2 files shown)


src/plugins/node/gc/nodegcplugin.cpp
```diff
@@ -149,7 +149,17 @@ void afterGC(v8::Isolate *isolate, GCType type, GCCallbackFlags flags) {
...
+	const char *gcType = NULL;
+        switch (type) {
+          case kGCTypeMarkSweepCompact: gcType = "M"; break;
+          case kGCTypeScavenge: gcType = "S"; break;
+#if NODE_VERSION_AT_LEAST(5, 0, 0)
+          case kGCTypeIncrementalMarking: gcType = "I"; break;
+          case kGCTypeProcessWeakCallbacks: gcType = "W"; break;
+#endif
+          // Should never happen, but call it minor if type is unrecognized.
+          default: gcType = "S"; break;
```
hhellyer: Can a GC always only be one type? The collector won't do two as one operation? The type values are designed to be masked together, that may be just for registration but I can't tell from the docs in `v8.h`.
As it stands if `type == kGCTypeMarkSweepCompact| kGCTypeProcessWeakCallbacks` the case statement would report gcType as `S` (`kGCTypeScavenge `).

sam-github: The pre-existing code here assumed it could only be one operation, as well as the new code in Node.js, but I'm looking through the v8 src now to confirm that is correct.

sam-github: Looking at the V8 code, it appears to me that the type is always a specific value, I think the code above is correct. @bnoordhuis you agree?

bnoordhuis: Yes, V8 always reports them individually.

hhellyer: Cool, I saw them being masked together and thought I'd better check.

hhellyer:
I've reverted this change as I missed that a test was failing on Travis with Node 8 where the new GC types exist as it hasn't been updated to expect the new output:

https://github.com/RuntimeTools/appmetrics/blob/master/tests/api_tests.js#L124

(I've also now discovered that github will let you revert a PR but won't let you re-open one. Sorry about that!)

sam-github:
Sorry, I thought I did `npm test` and it passed, I'm (supposed to be) on vacation today, I'll fix this Monday, its not at all urgent for the upcoming release.

PR #492, opened 2017-11-02
Add information for appmetrics-dash summary panel.

hhellyer:
- Add max used heap and total proportion of time in gc to gc messages.
- Add mean cpu usage stats.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,10 +222,13 @@ Requests an object containing all of the available environment information for t
 
 ### Event: 'cpu'
 Emitted when a CPU monitoring sample is taken.
+CPU times are reported as fractions of time used between 0.0 and 1.0.
 * `data` (Object) the data from the CPU sample:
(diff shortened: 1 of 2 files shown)


sjanuary:
@hhellyer - can the tests be updated to check for these new properties? See https://github.com/RuntimeTools/appmetrics/blob/master/tests/api_tests.js

tobespc:
typically, you process the data where you want to use it so doing the work in the dashboard would fit that model.  However, are there any other places (thinking of our current work) where we would also want to use this as part of feedback loop ?

sjanuary:
@tobespc if we need to write another engine to collect data that isn't part of the dashboard, then that probably wouldn't go into appmetrics base anyway. We could reconsider if needed.

sjanuary:
@hhellyer - can this be closed? I think the work was done in appmetrics-dash in the end?

stalleyj:
Closing, agreed with Howard.

Issue #494, opened 2017-11-02
Add a Couchbase Probe


astub:
I wanted to add a couchbase-probe to time some of the common bucket methods from [couchnode](https://github.com/couchbase/couchnode).

astub:
`'use strict';

var Probe = require('../lib/probe.js');
var aspect = require('../lib/aspect.js');
var util = require('util');
var am = require('../');

function CouchbaseProbe() {
  Probe.call(this, 'couchbase');
}
util.inherits(CouchbaseProbe, Probe);

CouchbaseProbe.prototype.aspectBucketMethod = function(bucket, method) {
  var that = this;

  aspect.before(bucket, method, function(target, methodName, methodArgs, probeData) {
    that.metricsProbeStart(probeData, method, methodArgs);
    if (aspect.findCallbackArg(methodArgs) != undefined) {
      aspect.aroundCallback(methodArgs, probeData, function(target, args) {
        that.metricsProbeEnd(probeData, method, methodArgs);
      });
    }
  });
};

// Most used couchbase bucket methods
const bucketMethods = ['upsert', 'insert', 'replace', 'remove', 'get', 'getMulti'];

CouchbaseProbe.prototype.attach = function(name, target) {
  var that = this;
  if (name != 'couchbase') return target;
  if (target.__ddProbeAttached__) return target;
  target.__ddProbeAttached__ = true;

  var mock = target['Mock']['Cluster'].prototype;
  var cluster = target['Cluster'].prototype;

  var data = {};
  aspect.after(mock, 'openBucket', data, function(target, methodName, args, probeData, bucket) {
    for(key in bucketMethods) {
      that.aspectBucketMethod(bucket, bucketMethods[key]);
    }
    return bucket;
  });

  aspect.after(cluster, 'openBucket', data, function(target, methodName, args, probeData, bucket) {
    for(key in bucketMethods) {
      that.aspectBucketMethod(bucket, bucketMethods[key]);
    }
    return bucket;
  });

  return target;
};

/*
 * Lightweight metrics probe for couchbase queries
 *
 * These provide:
 * 		time:		  time event started
 * 		bucket:		The bucket executed on
 *    method:   the method called on the bucket
 * 		duration:	the time for the request to respond
 */
CouchbaseProbe.prototype.metricsEnd = function(probeData, method, methodArgs) {
  if (probeData && probeData.timer) {
    probeData.timer.stop();
    var eventTimer = probeData.timer;
[...]

PR #498, opened 2017-11-03
added couchbase probe for couchnode package

astub:
#494


astub:
I'm not sure what is making codecov fail.

tobespc:
@astub before we can accept your change request, please can you read and accept 

https://github.com/RuntimeTools/appmetrics/blob/master/CONTRIBUTING.md

sjanuary:
@astub if your test can be run without needing a database to be set up (looks like a mock one?) then you can add it to the list of tests we run in package.json  https://github.com/RuntimeTools/appmetrics/blob/master/package.json#L32. We call out the tests individually because we have some probe tests that require a service setting up like a database so we don't run those by default. That would also fix the codecov decrease.

astub:
@sjanuary Hi. It uses couchbase.Mock so it does not need a couchbase server. I'm unsure what else i'm missing for this PR.

sjanuary:
@astub - you need to rebase as master has changed since this PR and there are conflicts. Also if you rename your test file to end in "tests.js" then it will be run automatically - see https://github.com/RuntimeTools/appmetrics/blob/master/package.json#L32

astub:
@sjanuary Thanks for your feedback. Anything else I need to improve/fix?

sjanuary:
@astub please could you also add a description the couchbase event to the Readme?  It needs to go in the table at the start and also [here](https://github.com/RuntimeTools/appmetrics/blob/master/README.md#api-dependency-events-probes).

astub:
closing this as its a copy of #506

PR #508, opened 2017-11-28
Fix double initialisation when two appmetrics instances are started (singleton bug)

sjanuary:
An issue was discovered with the implementation of the appmetrics global singleton.  Essentially a lot of initialization code was being run twice including attaching the probes and the socket.io probe doesn't have a global check for whether it is already attached.  This PR only runs the initialization code once.

See https://github.com/RuntimeTools/appmetrics-dash/issues/124

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,109 +14,9 @@
  * limitations under the License.
  *******************************************************************************/
 'use strict';
-var path = require('path');
-var main_filename = require.main != null ? require.main.filename : '';
-var module_dir = path.dirname(module.filename);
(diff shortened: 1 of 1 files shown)


a-roberts:
Confirmed this fixes a problem seen [here](https://github.com/RuntimeTools/appmetrics-dash/issues/1240), thanks @sjanuary for the quick fix!

gibfahn:
The ignore whitespace diff is much cleaner btw (for reviewers): https://github.com/RuntimeTools/appmetrics/pull/508/files?w=1

PR #515, opened 2018-01-09
Integrated Node.js DC into appmetrics.

shiyanf:
Integrated Node.js DC into appmetrics.
We have test this DC on the on-premise, bluemix and ICp on linux platform.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,7 @@ node_modules
 /agentcore.dll
 /heapdump.node
 headlesstestoutput*
+.nyc_output/
+nodejs_*.log
+plugins/
 
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,3 +2,6 @@
 [submodule "omr-agentcore"]
 	path = omr-agentcore
 	url = https://github.com/RuntimeTools/omr-agentcore
+[submodule "node-ibmapm"]
+	path = node-ibmapm
+	url = https://github.com/IBM-APM/node-ibmapm
--- a/index.js
+++ b/index.js
@@ -22,12 +22,12 @@ if (global.Appmetrics) {
   assert(
     global.Appmetrics.VERSION === VERSION,
     'Multiple versions of Node Application Metrics are being initialized.\n' +
-        'This version ' +
-        VERSION +
-        ' is incompatible with already initialized\n' +
-        'version ' +
-        global.Appmetrics.VERSION +
-        '.\n'
+      'This version ' +
+      VERSION +
+      ' is incompatible with already initialized\n' +
+      'version ' +
+      global.Appmetrics.VERSION +
+      '.\n'
   );
   exports = module.exports = global.Appmetrics;
 } else {
@@ -78,10 +78,10 @@ if (global.Appmetrics) {
     };
   }
   /*
-     * Load module probes into probes array by searching the probes directory.
-     * We handle the 'trace' probe as a special case because we don't want to put
-     * the probe hooks in by default due to the performance cost.
-     */
+ * Load module probes into probes array by searching the probes directory.
+ * We handle the 'trace' probe as a special case because we don't want to put
+ * the probe hooks in by default due to the performance cost.
+ */
   var probes = [];
   var traceProbe;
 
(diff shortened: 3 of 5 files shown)


tobespc:
@shiyanf whats the timescale for this work to be accepted ?  We need to do some testing on it and code reviews etc but this week we won't be able to focus on it due to other commitments

sjanuary:
I don't think I understand the motivation for this work.  What is it trying to achieve?  It seems to add a lot of code to appmetrics that is only used in a specific Bluemix scenario.  Also what are the performance implications?

As a side note, all the appmetrics tests are failing, but since I don't think we should be doing this at all that's probably not the main issue.

stalleyj:
Who would be responsible for maintaining/testing this function ? I think this would give a significant performance degradation to standard appmetrics. Legal work for the dependencies ?

sjanuary:
Also, in the current state, this adds a circular dependency to appmetrics, which means that appmetrics (potentially a different version) is installed as a dependency of itself and then instantiated twice. We absolutely cannot do this.

shiyanf:
Thanks for your comments, I will fix the code issues first.

shiyanf:
I updated the code according to above comments. 
1. Passed all of tests.
2. Remove the spaces.
3. Fixed a code issue of appmetrics, https://github.com/RuntimeTools/appmetrics/issues/516
4. About circular dependency to appmetrics, I have did some test and this will not cause problem. As Node.js DC need to run as a standalone module also, so it has to require 'appmetrics' in package.json.

shiyanf:
The codecov and codebeat for Node.js DC, you can find the badges in the https://github.com/IBM-APM/node-ibmapm.

sjanuary:
Thank you for the fix for #516.  Would you be able to provide this as a seperate pull request?  We like to have one PR for each separate issue/feature/bugfix, this is considered good practice in Git and allows us to review the 2 things separately.

Re. 4: We really definitely won't be able to accept a PR that adds appmetrics as a dependency of itself.  It might work in **some** scenarios but there are other scenarios in which it will cause the program to fail. It also nearly doubles the download size and build time for every appmetrics user and it's generally considered bad practice. Perhaps a 3rd module that depends on both appmetrics and IBM-APM would be a better solution?

shiyanf:
About #516 fix, as This fix in the master branch, it is hard to separate it now.

shiyanf:
About adds the "appmetrics" as a dependency, it is working on windows and linux, I have tested it. If you think there is any scenario will cause problem, please point it out. Thanks.

This definitely not doubles the download size, actually it almost same size as install the appmetrics and ibmapm(Node.js DC) separately.  

This definitely not doubles the build time also, the Node.js DC doesn't need any build.

The 3rd module is not a option now. The purpose of this PR is to integrated the Node.js DC into the appmetrics.

tobespc:
@shiyanf To be able for us to fully understand and accept these changes, we need to have each change as a separate PR.  I realise you have everything in 'master' but we still need these to be split out.  I'll setup a call so that we can discuss these changes in further detail

sjanuary:
FYI: To do work in a new git branch (assuming 'origin' points to your fork):
```
git remote add upstream https://github.com/RuntimeTools/appmetrics
git branch newfix
git checkout newfix
git fetch upstream
git reset --hard upstream/master
// make the code change and save it now
git add .
git commit -m "Fix for ABC"
git push origin newfix
```
You can then create a second pull request from the second branch.

https://softwareengineering.stackexchange.com/questions/11856/whats-wrong-with-circular-references discusses some general concerns around circular references and why it's considered bad design.

I should have been clearer that adding a circular dependency will double the build time and disk space for **appmetrics** itself, not necessarily the total including all other dependencies.  This is still not going to be acceptable to our users or the products that package appmetrics such as IBM SDK for Node.js or IBM API Connect.  The test and maintenance burden of such a complex setup is also a concern.

It would be good to have a call together like Toby suggested to explore alternative solutions.

gadekarl:
@tobespc Let's have a call and hash this out. You can go ahead and discuss with @shiyanf and team in your time zone

shiyanf:
I have separated the PR for the issue 516. https://github.com/RuntimeTools/appmetrics/pull/519
I will discuss the alternative solutions with Toby.

sjanuary:
@tobespc and @gadekarl have spoken on the phone and agreed to close this PR

PR #519, opened 2018-01-17
fix for the issue 516.

shiyanf:


sjanuary:
@shiyanf - thanks for the fix. I think this should also be fixed for HTTP outgoing as well as HTTPS, would you be able to add that too? The code is virtually identical - see https://github.com/RuntimeTools/appmetrics/blob/master/probes/http-outbound-probe.js#L127

shiyanf:
Done for http outbound probe also.

PR #522, opened 2018-03-07
Require ibmapm on first initialisaton of appmetrics.

hhellyer:


shiyanf:
I have created a new PR https://github.com/RuntimeTools/appmetrics/pull/523. It is just like this one. But will require a new "ibmapm-embed" module, which module has removed appmetrics from package.json. So  please close this. Thanks.

PR #523, opened 2018-03-21
Require ibmapm on first initialisaton of appmetrics.

shiyanf:
This code change is just like Howard's change https://github.com/hhellyer/appmetrics/commit/de7a636da225cd896fdaa8b1a4c7facd074647f9.
As we have create a new npm package ibmapm-embed, the ibmapm-embed has removed the appmetrics from package.json, and add a attach function. So change the require to "ibmapm-embed".

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,7 +32,6 @@ if (global.Appmetrics) {
   exports = module.exports = global.Appmetrics;
 } else {
   // This instance is the global, do all the setup here
-  global.Appmetrics = module.exports;
   module.exports.VERSION = VERSION;
 
   var path = require('path');
@@ -380,4 +379,7 @@ if (global.Appmetrics) {
     return this;
   };
 
+  // Expose global instance when appmetrics is fully initialised.
+  global.Appmetrics = module.exports;
+  require('ibmapm-embed');
 }
--- a/index.js
+++ b/index.js
@@ -32,6 +32,7 @@ if (global.Appmetrics) {
   exports = module.exports = global.Appmetrics;
 } else {
   // This instance is the global, do all the setup here
+  global.Appmetrics = module.exports;
   module.exports.VERSION = VERSION;
 
   var path = require('path');
@@ -378,8 +379,6 @@ if (global.Appmetrics) {
 
     return this;
   };
-
-  // Expose global instance when appmetrics is fully initialised.
-  global.Appmetrics = module.exports;
+  //Require the Node.js DC, when appmetrics is fully initialised.
   require('ibmapm-embed');
 }
--- a/package.json
+++ b/package.json
@@ -9,7 +9,8 @@
     "nan": "2.x",
     "tar": "2.x",
     "semver": "^5.3.0",
-    "jszip": "2.5.x"
+    "jszip": "2.5.x",
+    "ibmapm-embed": "^1.0.0"
   },
   "bundleDependencies": [
     "tar"


sjanuary:
This looks ok from the appmetrics side although I'm not sure why `global.Appmetrics = module.exports;` gets moved to the bottom? It is assigning the same object so I don't think that should be needed.

I understand that the agreement is that ibm-apm will do nothing and have zero impact unless it finds a certain environment variable set, however I see that loggers are set up and it attempts to open files (see console output below from our test suite) and I don't think this is quite right.  Looking at https://github.com/IBM-APM/node-ibmapm-embed/blob/master/index.js I would expect to see a check and `return` right near the top of the file, even before most of the require statements.

```
    [2018-03-21 03:42:04.076] [INFO] knj_log - KNJ_LOG_LEVEL is not set or not set correctly through environment variables.
    [2018-03-21 03:42:04.080] [INFO] knj_log - The program set default log level to INFO.
    [2018-03-21 03:42:04.379] [INFO] util.js - ENOENT: no such file or directory, open '/home/travis/build/RuntimeTools/appmetrics/tests/probes//package.json'
    [2018-03-21 03:42:04.379] [INFO] util.js - ENOENT: no such file or directory, open '/home/travis/build/RuntimeTools/appmetrics/tests//package.json'
    [2018-03-21 03:42:04.380] [INFO] BAMPlugin.js -  VCAP_APPLICATION env variable is not found.
```

shiyanf:
@sjanuary 
I have clean the DC code, to remove some unnecessary logic and logs, and updated the npm package also. You can have a review again.
The Node.js DC cannot return just by checking a environment, except  "process.env.NODEJS_DC_DISABLE" or "process.env.ITCAM_DC_ENABLED" is set. We have such logic in the "https://github.com/IBM-APM/node-ibmapm-embed/blob/master/index.js" now.
The DC need to check whether the APM's server URL is set or not.  Most time the DC will return directly, if there is no APM URL is set, the DC will not start. But in the ICp case, the user can set the APM URL by mount a kubernetes secrets, the DC will polling the user's secret( mount into the container)  by every minute.  You can refer to https://github.com/IBM-APM/node-ibmapm-embed/blob/master/lib/ibm_apm_restclient/lib/plugins/BAMPlugin.js .

> if (k8sutil.isICP() && !process.env.IBAM_INGRESS_URL) {
        intervalObjSec = setInterval(pollMountSecret, 60000);
        intervalObjSec.unref();
    }

tobespc:
I would like to confirm the expected behaviour here and how we expect the code to work.  Because we will be requiring you, we need to make sure that in all the environments where BAMDC is not running, the impact of calling your code is as minimal as possible.  This is why we want your code to return straight away, without setting up logs, variables or anything else that is not required if BAMDC is not going to be used.   Can you have the environment checks for {process.env.NODEJS_DC_DISABLE" or "process.env.ITCAM_DC_ENABLED} as the very first thing your code does and then return ?

shiyanf:
@tobespc I have moved the env checking to the top of index.js, and check it into git repo. will update to npm  server once the review finished. Thanks.

tobespc:
ok, @sjanuary and myself and some more concerns.   How are those values going to be set in the default case.  The default case would be that neither of those values are set and this would mean you would not return to us.  We would need the behaviour to be always return UNLESS the values were enabled by BAM.  Any environment variables could be used but we have to assume that the default case is users are not using BAM and have not set any env variables to say "do not use BAM"

sjanuary:
@tobespc as you know I have 2 weeks vacation coming up after today so someone else may need to merge this but what I am looking for before merging is to see no APM output in the tests (currently we see the 5 lines I quoted above e.g. https://travis-ci.org/RuntimeTools/appmetrics/jobs/356180510).  This will help us confirm that APM is exiting immediately as well as inspecting https://github.com/IBM-APM/node-ibmapm-embed/blob/master/index.js. Once the new ibmapm-embed package has been published we can manually request a rerun of the Travis build for this PR to see the results, they won't run automatically unless the content of the PR changes.

I was also looking for the line `global.Appmetrics = module.exports;` to not be moved, I don't think moving it makes any functional difference so it would make sense to leave it as it is unless I'm missing something.  This may be overly picky but I think the code is easier to read to leave it straight after the else statement.

@shiyanf Will ibmapm-embed conform to semver (https://semver.org/)?  So no breaking changes in minor and patch releases?

shiyanf:
@tobespc @sjanuary 

- I rollbacked the `global.Appmetrics = module.exports;` change.  The node.js DC conform to semver, there is no breaking changes in minor and patch releases.

shiyanf:
@tobespc @sjanuary About the node.js DC's behavior in default case. As I commented before, the node.js DC will check if a APM server URL is set, not by checking a environment variable. Unless the node.js DC find APM server URL, otherwise the Node.js DC will not start at all. 
It is not a good idea to enable/disable Node.js DC by set a environment. Actually even the user require node.js dc directly, they only need to update package.json and 'app.js'.

qiyaod:
@tobespc @sjanuary APM node.js DC has some usability design:
 1. Pulling a file (in Kub environment, it can be a mounted ConfigMap/Secret) to get the APM server connection info.
2. Auto-discover the APM server connection info.

These design allow the customers get their deployed application monitored by APM automatically without re-start the application, or switch to a new APM server.

Your suggestion of "set env variables to say "use BAM" or "do not use BAM"" will "kill" this usability design, because the user has to re-start/re-deploy their app anyway to switch the env variable for enable/disable.

To have the benefit of APM usability design, I think we have 2 options:

1) Keep the "pulling" and "auto-disvocer" logic in APM node.js DC. APM node.js DC will be always enabled. While APM node.js will keep minimal load ("pulling", "auto-disvocer" logic) before detecting valid APM server connection info.
2) Move the APM server connection info "pulling" and "auto-disvocer" logic to appmetrics, appmetrics can enable/disable the APM node.js DC base on if valid APM server connection info discovered.

I prefer option 1), which requires less code changes and test. Do you agree?

hhellyer:
When we originally discussed enabling APM the late enablement option in kubernetes didn't come up.

I can see why you are keen for the late enablement inside kubernetes, I'm not sure how much overhead there will be to regularly checking whether the secret you use has been added. However I think there may be a bigger problem - appmetrics can only instrument npm's as they are required. If APM starts appmetrics well after application startup then we won't instrument that code and won't be monitoring correctly. Not restarting the application when enabling appmetrics may mean you lose data, depending on what you are using this may be a major problem.

Appmetrics is designed to be required before any npm's that need to be monitored:
https://github.com/RuntimeTools/appmetrics#modifying-your-application-to-use-appmetrics

Users requiring appmetrics too late is one of the number one causes of issues when using appmetrics. Looking at  it closely the code in node-ibmapm-embed is likely to be requiring too many things before it loads appmetrics (if it wasn't already loaded).

Restarting an application to enable appmetrics monitoring is a requirement, I think the late enablement option is likely to result in you receiving incomplete data and the user seeing strange behaviour. The environment variable option is probably better as late enabling appmetrics (via APM or otherwise) isn't supported.

shiyanf:
@hhellyer the node-ibmapm-embed is only used for the appmetrics, this one cannot be used in the user's app by 'require('ibmapm-embed')'. To use this node-ibmapm-embed, the user just need to add 'require('appmetrics')' in the app.  So when the node-ibmapm-embed is required the appmetrics should be loaded already.

sjanuary:
@qiyaod @shiyanf could you please point us to the code responsible for both the "pulling" and "autodiscovery" routes to enabling APM so that we can better understand what is happening and what the overhead will be?  I agree that it makes sense to keep the code in APM but we would like to understand the impact on appmetrics.  I also think that making it as minimal as possible should mean no logging output, not because this is a performance issue but because it could be confusing to appmetrics users who are not using APM.

shiyanf:
@sjanuary In the https://github.com/IBM-APM/node-ibmapm-embed/blob/master/lib/ibm_apm_restclient/lib/plugins/BAMPlugin.js , the function pollMountSecret will be invoked every minute.
```
    if (k8sutil.isICP() && !process.env.IBAM_INGRESS_URL) {
        intervalObjSec = setInterval(pollMountSecret, 60000);
        intervalObjSec.unref();
    }
```
And in the function `poollMountSecret`, once a server url is setted, a boolean `initialized ` is set true also. Then a event `bam_initialized` will be emitted,  

```
function emitEvent() {
    if (initialized) {
        logger.debug('Emit bam_initialized event. count = ', count, initialized);
        ready_event.emit('bam_initialized');
        if (count > 10) {
            clearInterval(intervalObj);
        }
        count++;
    }
}
```
And the event will trigger the function `startDC()` in the https://github.com/IBM-APM/node-ibmapm-embed/blob/master/index.js , then the DC is started.

qiyaod:
@shiyanf can you provide the log of the "no APM" scenario, let's make sure if no APM is setting the log is not confusing to the appmetrics users who are not using APM.

shiyanf:
If there is no APM server, the Node.js DC will not dump any log to console and file.
```
D:\projectdoc\node.js\testware\acmeair-nodejs>node app.js
[Tue Apr 17 15:23:25 2018] com.ibm.diagnostics.healthcenter.loader INFO: Node Application Metrics 4.0.0.201804171520 (Agent Core 3.2.6)
[2018-04-17 15:23:26.497] [INFO] app - host:port==localhost:9080
[2018-04-17 15:23:26.499] [INFO] app - db type==mongo
[2018-04-17 15:23:26.553] [INFO] routes - Use dataaccess:../dataaccess/mongo/index.js
[2018-04-17 15:23:26.821] [INFO] dataaccess/mongo - mongo: "mongodb://9.42.26.112:27017/"
[2018-04-17 15:23:28.138] [INFO] app - Initialized database connections
[2018-04-17 15:23:28.143] [INFO] app - Express server listening on port 9080
[2018-04-17 15:23:28.565] [INFO] dataaccess/mongo - ensureIndex:null:customerId_1
[2018-04-17 15:23:28.566] [INFO] dataaccess/mongo - ensureIndex:null:originPort_1_destPort_2
[2018-04-17 15:23:28.568] [INFO] dataaccess/mongo - ensureIndex:null:flightSegmentId_1_scheduledDepartureTime_2
```

sjanuary:
logs now look clean

qiyaod:
@sjanuary so your concerns on "APM impact and log message" are addressed, correct? When will you merge the code, then we can verify and close this? This has been quite a while, let's get it close soon and move on.

sjanuary:
@stalleyj should we do a quick performance/memory footprint comparison before merging this?

tobespc:
@qiyaod Can you provide some details on the performance impact please.  I know this has been going on for a long time but we have to consider some of the other places that appmetrics is deployed.  For example, Node-Red ship with us and we have done a lot of work to make sure our footprint in terms of disk space and cpu are as minimal as possible.  What I would like to know is what is the disk footprint of installing appmetrics with your code now included?   I'm going to assume for now that we would have no additional cpu overhead if running in an environment where bam is not available.  So this would mean the only impact to existing appmetrics users would be disk usage as additional packages are now going to be installed.

If you could provide this information asap, we can then look to close this issue out providing the increase is acceptable

shiyanf:
@tobespc This is the performance test for Appmetrics and Appmetrics with Node.js DC disabled.
5d1b9556e848499

For disk footprint, because there are some duplicated opensource lib in ibmapm and appmetrics, and ibmapm seems require a more lower version, then it will be smaller than what appmetrics require. e.g. the following highlight, semver in ibmapm integrated is much more smaller than semver in pure appmetrics:
249407989c506a56

qiyaod:
@tobespc I think @shiyanf means the APM Node.js DC is enabled but no APM server is set, so Node.js DC just keep polling the APM server settings, not collecting and sending data.

qiyaod:
@tobespc In the table, "pure" is the baseline that means APM DC is disabled completely, "integrate" means APM DC is enabled but no APM server is discovered.

qiyaod:
@tobespc BTW, the "responsetime" is in ms.

yuecchen:
@sjanuary it requires half an hour to run one round, and we test 4 rounds for each(pure appmetrics and integrated appmetrics) it requires at least 4hours not counting the operation between rounds. To test 10 round is too large work and we really don't have more time to do it.   
Good news is: recheck our raw data, we found 2 test data are not very reasonable, after we delete them and do recalculate, we found the resptime downgrade become 0.81%. 

image

From this screen shot, the "delta" is what we do recalculate, the table with frame is the original calculation.

Also, as resptime is in a "ms" unit, the value is already very small, so every little environment wave, we shake it to 2% or larger. the following screenshot will show that the test result between each round is also big. and our test result basically within the margin of error. 

image
 
To make sure the test result, we delete the "shake" one and recalculate the downgrade to 0.81%

qiyaod:
@sjanuary in summary: 1) The response time is in "ms", any small environment variation will be magnified by percentage number. 2) From Yue several  rounds of testing, even the "pure" (appmetrics only without APM node.js DC), the response time variation can be up to 2.54%. 3) So we think in the original report the 2% is in acceptable variation scope. Any further concern?

sjanuary:
I've spoken to Toby this morning and we've agreed that the slight difference between response time measurements aren't an issue for merging. Going to pull this in now.

shiyanf:
@sjanuary Thanks for your great help! Could you please tell me when this code change can be published to https://www.npmjs.com/package/appmetrics ?

sjanuary:
@shiyanf You're welcome, thanks for making all the changes we requested. We'd like to get this released soon but there are some internal build machine issues currently which are blocking us.  We are discussing alternative options and we will let you know as soon as we are in a position to release.

qiyaod:
@sjanuary did you fix your build machine issues and have a latest appmetrics release for verification?

sjanuary:
We're working on it, should be sometime this week.

qiyaod:
@sjanuary @tobespc any progress on this? Thanks!

tobespc:
we have pushed back to you on legal requirements.  There is an email chain sent by Sian that details the packages that APM are using that we don't yet have clearance for

sjanuary:
@qiyaod the build is ready but as Toby said we are waiting for a reply to the email I sent to you last week about dependencies. We should take that discussion offline.

qiyaod:
@sjanuary our PM is looking at the Pedigree Review and sent you a mail already. So the new build will be publish to NPM after that closed, correct?

sjanuary:
@qiyaod now published to NPM

PR #532, opened 2018-07-17
Add LTS and Commercial Support information to README.md

seabaylea:
Add badges and information for LTS policy and commercial support.


sjanuary:
@seabaylea - http://github.com/CloudNativeJS/ModuleLT seems to be an empty repo at the moment, I assume that's going to have content added soon?

Tests seem to have failed because default Node on Travis has moved up from 8.x to 10.x, which we don't yet support.  I will create a separate PR to change the Travis build to just build on 6 and 8, although I guess we should also start supporting Node 10 fairly soon.

seabaylea:
Yes, that's expected - the repo will be populated in the next couple of days.

sjanuary:
Ok, thanks. I've fixed the tests on master, no point rebasing as this PR only changes the README.

PR #534, opened 2018-07-17
coverage: exclude directories

BethGriggs:


sjanuary:
@BethGriggs why do you want to exclude these directories?  `lib` and `probes` contain a good chunk of the actual code for appmetrics, `headless_zip.js` is also appmetrics code.

Issue #537, opened 2018-08-24
Allow For IBM i PASE Support


ab-m:
Currently Appmetrics has issues when installed on the IBM i PASE environment.

When using Node verison 8.X appmetrics can be npm installed but there are runtime errors:

`Error: rtld: 0712-001 Symbol _ZN2v816FunctionTemplate3NewEPNS_7IsolateEPFvRKNS_20FunctionCallbackInfoINS_5ValueEEEENS_5LocalIS4_EENSA_INS_9SignatureEEEiNS_19ConstructorBehaviorE was referenced
      from module /appmetrics-test/node_modules/appmetrics/appmetrics.node(), but a runtime definition
      of the symbol was not found.
rtld: 0712-002 fatal error: exiting.`

When using Node version 10.X 
- currently there are no prebuilt binaries so the installation falls back to node-gyp rebuild. During compilation of the omr-agentcore sub-module errors occur:

`../omr-agentcore/src/ibmras/monitoring/plugins/common/environment/envplugin.cpp:364:66: error: 'getargs' was not declared in this scope`

I am working on PR that fixes this run time Error but there were also other issues that needed to be addressed inside the omr-agentcore sub-module.

Some changes were needed to properly include and use libraries available on PASE. One of these issues is present above during the Node version 10.X install.

I am also working on a separate PR for the omr-agentcore repository , and have posted an issue [there](https://github.com/RuntimeTools/omr-agentcore/issues/68). 

I would like to have both reviewed and if found to be OK have this appmetrics repository use the new version of omr-agentcore (with the changes from the PR) as the sub-module.

That way appmetrics can be installed and used properly on the IBM i.

mattcolegate:
@abmusse Change has been merged - can you please test and see if this issue can be closed?

PR #538, opened 2018-08-27
Adding support for IBM i PASE installs

ab-m:
Pull Request in response to issue #537

### DIFF
--- a/src/appmetrics.cpp
+++ b/src/appmetrics.cpp
@@ -285,7 +285,7 @@ static void* getMonitorApiFunction(std::string pluginPath, std::string functionN
     std::string libname = "hcapiplugin.dll";
 #elif defined(__MACH__) || defined(__APPLE__)
     std::string libname = "libhcapiplugin.dylib";
-#elif defined (__AIX__) || defined(AIX)
+#elif defined (__AIX__) || defined(_AIX)
     std::string libname = "libhcapiplugin.a";
 #else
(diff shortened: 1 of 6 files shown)


src/appmetrics.cpp
```diff
@@ -285,7 +285,7 @@ static void* getMonitorApiFunction(std::string pluginPath, std::string functionN
     std::string libname = "hcapiplugin.dll";
 #elif defined(__MACH__) || defined(__APPLE__)
     std::string libname = "libhcapiplugin.dylib";
-#elif defined (__AIX__) || defined(AIX)
```
sjanuary: @ab-m - what's the purpose of this change?

ab-m: @sjanuary `_AIX` is predefined on both AIX & IBM i PASE systems whereas `AIX` is not predefined on. In this instance both AIX & PASE would follow the same code path so `_AIX` was used.

Alternatively another `|| defined(__PASE__)` could have been appended to the line instead but the first approach using `_AIX` seemed cleaner.

sjanuary: that's fine, I just wanted to understand the reason for it

sjanuary:
@ab-m I have tagged a new omr-agentcore release so you can now update the submodule in your branch to point to that tag.  Also if you could add something to the README about iPASE that would be great, i.e. what works and what doesn't work.

PR #541, opened 2018-09-02
mysql2 probe

realhidden:
Adds a new mysql2 probe (for package https://www.npmjs.com/package/mysql2), the mysql2 package is compatible with the original mysql one, so I used the same code and emitting 'mysql' data to be 100% compatible with the original probe and the appmetrics-dash.

### DIFF
--- a/probes/mysql2-probe.js
+++ b/probes/mysql2-probe.js
@@ -1,0 +1,87 @@
+/*******************************************************************************
+ * Copyright 2015 IBM Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


sjanuary:
@realhidden - thank you for the contribution. Please read https://github.com/RuntimeTools/appmetrics/blob/master/CONTRIBUTING.md for instructions on how to contribute to appmetrics - your first PR needs to add your name into AUTHORS.md as an indication that you agree to the Contributor Licence Agreement.

It doesn't seem like we have a test for the mysql probe at the moment, would you be interested in adding one for this new probe?  Tests under https://github.com/RuntimeTools/appmetrics/tree/master/tests/probes are not run by default if they require something like an external database but they are useful when writing new probes, testing changes to a probe and fixing bugs.

PR #552, opened 2018-11-05
Fix travis config to use Node 10 and update TAR dependency 

sjanuary:


sjanuary:
This got checked in with #553

PR #553, opened 2018-11-06
Update agentcore to fix ZOS build problem

sjanuary:


sjanuary:
Thanks @mattcolegate, merging

PR #562, opened 2018-11-29
Added extra information when profiling mongodb

sebastianarena:
Very useful information comes in the second parameter, such as sort, limit and projection


lardissone (approved):
Important to have `sort` information in output.

PR #602, opened 2019-07-05
update ibmapm-embed version

yuecchen:


yuecchen:
@tobespc Could you please help integrate this code change into the appmetrics release version?

yuecchen:
> why not make the version ^19.6.0 rather than >=19.6.0 ?

We updated our version as following format: <year>.<month>.<fix>
If I use `^`, I still need to update the package.json next year.

